Issue #17, opened 2026-03-06
feat(testing): missing forward-pass coverage for Horn degeneracy and 1D batches


hunter-heidenreich:
### 1. No Horn forward-pass degeneracy tests

`TestDegeneracy` only tests Kabsch/Umeyama origin collapse. There is no equivalent test for Horn.

`TestHornDifferentiabilityTraps` (in `test_differentiability_traps.py`) covers gradient stability under degeneracy, but does not assert anything about the *values* returned -- only that gradients are finite.

Missing: a `TestHornDegeneracy` class (or extension to `TestDegeneracy`) that checks:
- Origin collapse (`P = 0` or `Q = 0`): rotation should be identity, scale should be finite.
- Collinear inputs: rotation should be a valid proper rotation (`det == 1`, `R @ R.T == I`).
- Coplanar inputs: same.

### 2. No 1D batch forward-pass test for Kabsch/Umeyama

`test_forward_pass_matches_sequential_computation_when_nd_batched` tests a `[2, 3, N, D]` batch. There is no equivalent test for a plain `[B, N, D]` batch.

The `batch_points` fixture (`[5, N, D]`) is already used in `test_gradient_verification.py` for gradient correctness, but the forward-pass output values (R, t, rmsd) are never verified batch-vs-sequential for the 1D batch shape. A regression in the batch broadcasting logic for `[B, N, D]` would not be caught by forward-pass tests.

Issue #15, opened 2026-03-06
fix(testing): dead code and no-op expressions in test suite


hunter-heidenreich:
### 1. Duplicated `_compute_sequential_expected_tensors`

The function is defined identically in two places:
- `test_forward_pass_equivalence.py:15-53`
- `utils.py:146-184`

The copy in `test_forward_pass_equivalence.py` is never imported from `utils.py` -- it is a verbatim duplicate. One should be removed; the test file should import from `utils`.

### 2. Dead expressions in `test_differentiability_traps.py`

`P_np.shape[-1]` is evaluated but the result is discarded on at least 7 lines (22, 30, 46, 92, 119, 144, 255). These look like unfinished refactors where `dim = P_np.shape[-1]` was intended. Since `dim` is never used in those methods, the line can simply be removed.

### 3. Always-true `if` branch in `test_degeneracy.py`

```python
# test_degeneracy.py:44
if collapse_target in ["P", "Q", "Both"]:
    assert np.isfinite(c_res)
```

`collapse_target` is parametrized exclusively as `["P", "Q", "Both"]`, so the guard is always `True`. The `if` is dead and should be removed, leaving the assertion unconditional. If a fourth `collapse_target` case is ever intended (e.g. `"Neither"`), the branch should be explicitly documented.

Issue #13, opened 2026-03-06
feat(testing): convert fixed-seed gradient tests to Hypothesis-driven


hunter-heidenreich:
## Problem

All finite-difference gradient verification tests in `test_gradient_verification.py` use fixed NumPy seeds:

```python
np.random.seed(42)   # test_gradients_match_finite_differences_when_perturbed
np.random.seed(123)  # test_gradients_match_finite_differences_when_purely_random
```

These tests run against exactly the same two point clouds on every CI run. A correctness regression in the autograd backward pass for an unusual input shape or value range would not be caught until a user reports it.

## Requested change

Replace at least the 3D variants with `@given(point_clouds_3d())` tests, using `assume` to reject degenerate cases where finite differences are unreliable (e.g. near-zero singular values):

```python
@given(st.tuples(point_clouds_3d(), point_clouds_3d()))
@settings(max_examples=50, suppress_health_check=[HealthCheck.too_slow], deadline=None)
def test_gradients_match_finite_differences_hypothesis(self, PQ, adapter, algo, wrt):
    P_np, Q_np = PQ
    assume(np.linalg.matrix_rank(P_np - P_np.mean(0)) == P_np.shape[-1])
    ...
```

The ND variants can remain fixed-seed for now given the multi-framework overhead, but the 3D case is cheap enough to drive with Hypothesis.

## Scope

- `TestGradientVerification.test_gradients_match_finite_differences_when_perturbed` (3D case)
- `TestHornGradientVerification.test_gradients_match_finite_differences`
- `TestHornGradientVerification.test_gradients_purely_random`

Issue #12, opened 2026-03-06
feat(testing): add Hypothesis strategies for degenerate and batched inputs


hunter-heidenreich:
## Problem

`test_differentiability_traps.py` and `test_degeneracy.py` use fixed hardcoded arrays (e.g. `np.zeros((5, 3))`, a single collinear arrangement). Hypothesis is most powerful here -- shrinking, the example database, and diverse coverage -- but it is not used for any degeneracy scenario.

Similarly, there are no strategies for batched inputs, so the batch-vs-sequential consistency check in `test_gradient_verification.py` uses only the single hardcoded fixture from `conftest.py`.

### Degenerate input strategies (in `strategies.py`)

```python
@st.composite
def nearly_collinear_3d(draw):
    """Points near a 1-D line with small Gaussian noise."""

@st.composite
def nearly_coplanar_nd(draw, dim=None):
    """Points near a (dim-1) hyperplane with small noise."""

@st.composite
def near_duplicate_cloud(draw):
    """Two point clouds where most points nearly coincide."""

@st.composite
def extreme_scale_cloud(draw):
    """Point clouds with very large or very small coordinate magnitudes."""
```

### Batched input strategy

```python
@st.composite
def batched_point_clouds(draw, batch_dims=None, dim=None):
    """Arbitrary leading batch dims + [N, D] point cloud."""
```

## Why this matters

- The `SafeSVD` / `SafeEigh` gradient guards are the core numerical innovation of this library. Testing them only on fixed inputs means Hypothesis's shrinking cannot produce minimal failing examples if a new degenerate case is introduced.
- Extreme-scale inputs can expose float16/bfloat16 overflow in Umeyama (currently only checked via `pytest.skip` heuristics).

## Scope

Add strategies; wire them into at least one `@given` test per degeneracy class in `test_differentiability_traps.py` (gradient stability) and `test_degeneracy.py` (forward-pass correctness).

PR #97, opened 2026-03-08
test: expand test suite coverage, invariants, Hypothesis strategies, and fixture correctness

### LINKED ISSUE #11: feat(testing): add missing invariant property tests
## Missing invariant properties

The following mathematical invariants of Kabsch/Horn are not covered by any Hypothesis-driven property test. Each should hold for all valid random inputs and is a good candidate for a `@given`-based test in `test_properties.py`.

### 1. Roundtrip / residual consistency

If `kabsch(P, Q) = (R, t, rmsd)`, then:

```
||P @ R.T + t - Q||_F / sqrt(N)  ==  rmsd
```

This property verifies that the returned `rmsd` is actually the residual of the returned transform -- not just a separately computed value.

### 2. Inversion / commutativity

```
kabsch(P, Q) = (R, t, rmsd)  =>  kabsch(Q, P) = (R.T, -R @ t, rmsd)
```

The RMSD should be the same in both directions; the rotation and translation should be exact inverses.

### 3. Rigid-body frame invariance of RMSD

For any proper rotation `S` and translation `u`:

```
rmsd(S @ P.T).T + u, (S @ Q.T).T + u)  ==  rmsd(P, Q)
```

The RMSD (and recovered R up to conjugation) should be invariant to simultaneous rigid-body transformation of both inputs.

### 4. Translation invariance of R

Shifting both `P` and `Q` by the same vector should not change the recovered rotation `R`.

### 5. Scale invariance of R for Kabsch

Scaling both `P` and `Q` by the same positive scalar should not change `R` (only RMSD and translation scale).

### 6. Umeyama / Horn-with-scale: exact scale recovery

When `Q = c * (P @ R.T) + t` exactly, `kabsch_umeyama(P, Q)` should recover `c`. This complements the existing rotation recovery test but is not currently tested.

## Suggested structure

Add a `TestAlignmentInvariants` class (or extend `TestRotationInvariants`) in `test_properties.py`. Use `aligned_pair_nd` / `aligned_pair_3d` as the input strategy where a known transform is needed. Cover numpy at minimum; cross-framework coverage is a bonus.

hunter-heidenreich:
## Summary

Expands the test suite with mathematical invariant tests, Hypothesis-driven property checks, new degeneracy coverage, double backward tests, and fixes to pre-existing gaps in fixture correctness and forward-pass validation.

Closes #11 #12 #13 #15 #17.

### New test coverage

- **Mathematical invariants** (`test_properties.py`): `TestAlignmentInvariants` with 7 properties -- RMSD/residual consistency, RMSD symmetry, rigid-transform invariance, translation invariance, uniform scale invariance, exact scale recovery for Umeyama and Horn
- **Hypothesis strategies** (`strategies.py`): `nearly_collinear_3d`, `nearly_coplanar_nd`, `extreme_scale_cloud`; used in degeneracy and gradient trap tests
- **Degeneracy coverage** (`test_degeneracy.py`): `TestHornDegeneracy` (origin-collapse) and `TestHornDegeneracyGeometric` (collinear/coplanar inputs for Horn)
- **Gradient variety** (`test_gradient_verification.py`): `test_gradients_match_finite_differences_hypothesis` for both Kabsch and Horn; `test_gradients_stable_nearly_collinear_different_clouds` (independent near-collinear P and Q); `test_safe_svd_gradient_reduces_rmsd_at_hypothesis_near_degenerate` (descent-direction check at near-degenerate inputs)
- **Double backward** (`test_gradient_verification.py`): float32 and float64 double-backward coverage for JAX, TensorFlow, and MLX
- **Source-of-truth tests** (`test_properties.py`): `test_umeyama_equals_kabsch_when_no_scale_change`, `test_rotation_is_not_unique_when_cross_covariance_is_degenerate`, cross-framework `TestKabschRecoveryND`
- **1D-batch equivalence** (`test_forward_pass_equivalence.py`): `test_forward_pass_matches_sequential_computation_when_batched` for both Kabsch/Umeyama and Horn

### Fixture and infrastructure correctness

- **Batch fixtures** (`conftest.py`): `batch_points`, `nd_batch_points`, `horn_batch_points`, `horn_nd_batch_points` now apply an independent random rotation + translation per batch element; pure translation offsets meant batch consistency tests could not catch batched-rotation bugs
- **`c_true` variation** (`conftest.py`): drawn from `rng.uniform(0.5, 5.0)` instead of hardcoded `2.5` so scale recovery is verified at a non-trivial value each run
- **RMSD silently skipped**: `test_returns_correct_transform_when_points_have_known_transformation` and `test_returns_correct_transform_for_known_input` now assert `rmsd_expected=0.0` -- Q is an exact transform of P so RMSD must be zero
- **Reconstruction test**: `test_reconstruction_aligns_point_clouds` added to both `TestForwardPassEquivalence` and `TestHornForwardPassEquivalence`; the only cross-framework check that `c*(P@R.T)+t ≈ Q` holds directly, which is the property users depend on most
- **Rotation validity tests** (`test_properties.py`): all four orthogonality/det tests and `test_no_rotation_achieves_lower_rmsd` switched from translation-only Q (R ≈ I trivially) to `aligned_pair_nd()`/`aligned_pair_3d()` so the non-identity rotation path is actually exercised
- **`compute_sequential_expected_tensors`** (`utils.py`): generalized from a hard-coded 2-level loop to `np.ndindex` so it supports arbitrary leading batch dims
- **`JAXAdapter.convert_out`** (`adapters.py`): bfloat16 and float16 are cast before `np.array()`, matching MLXAdapter and preventing void/object dtype on some JAX versions

### Hypothesis correctness fixes

- Decorator ordering: corrected `@given`/`@settings` vs `@pytest.mark.parametrize` ordering across 12 affected methods
- `assume()` on parametrized adapter args replaced with `pytest.skip`; data-dependent dim guards kept as `assume()`
- Strategies use direct `draw()` rather than seeded numpy RNG inside composites
- `_paired_clouds_nd` flatmap converted to `@st.composite`; `_paired_with_shift` draws shift `v` in the same composite
- Dead strategies removed: `near_duplicate_cloud`, `batched_point_clouds`
- `nearly_coplanar_nd` draws `dim >= 3` (coplanar is undefined for dim=2)
- `test_gradients_match_finite_differences_hypothesis` skips float32 (atol*50 = 2.5 is vacuous for a gradient check)
- Float16/bfloat16 skips added to `test_gradients_stable_nearly_coplanar_hypothesis` (umeyama) and widened in `test_gradients_stable_nearly_collinear_different_clouds` (all algos); both tests overflow float16 gradient range on the affected inputs

### README

Added a **"Verified Properties"** section covering output invariants, correctness invariants, cross-algorithm consistency, gradient stability, and known algorithm boundaries -- each claim cites the specific test that proves it. Updated the "Testing" section from a vague category list to a per-file table, and updated "Stabilizing Gradients" to reference the backing test files.

## Test plan

- [x] `uv run pytest tests/test_forward_pass_equivalence.py tests/test_properties.py -q` -- 1277 passed
- [x] `uv run pytest tests/ -q` -- 6763 passed, 408 skipped, 4 xfailed
- [x] `uv run ruff check . && uv run ruff format --check .` -- all checks passed


hunter-heidenreich:
## Test Architecture Review

The batch-fixture redesign (`batch_points`, `nd_batch_points`, `horn_batch_points`, `horn_nd_batch_points`) is the most architecturally significant change and is done correctly -- independent rotations per element let the batch tests actually exercise batched-rotation codepaths. The `_get_random_rotation` helper is correctly centralised in `conftest.py` and reused by all fixtures. The `JAXAdapter.convert_out` fix is correct and well-targeted.

### Issues

#### [RECOMMENDED] `TFAdapter.convert_out` not updated alongside `JAXAdapter`

`adapters.py:295-299` -- the PR's stated goal for the `JAXAdapter` fix is "bfloat16 and float16 are cast before `np.array()`, matching MLXAdapter". The fix was correctly applied to `JAXAdapter`:

```python
# JAXAdapter (fixed in this PR)
if obj.dtype in (jnp.bfloat16, jnp.float16):
    obj = obj.astype(jnp.float32)
return np.array(obj)
```

But `TFAdapter.convert_out` was left with float16-only handling:

```python
# TFAdapter (unchanged)
ret = obj.numpy()
if ret.dtype in (np.float16,):
    ret = ret.astype(np.float32)
```

TF bfloat16 maps to a custom numpy dtype that participates in comparisons correctly today, so this may not cause failures currently. However, it breaks the architectural invariant that all adapters cast low-precision dtypes before returning numpy arrays. If a future test asserts dtype equality or does arithmetic on the raw numpy array, TF-bfloat16 will behave differently from the other three adapters. Apply the same pre-cast pattern to `TFAdapter.convert_out`.

#### [NITPICK] `TestDegeneracy` manually dispatches instead of using `get_transform_func()`

`test_degeneracy.py:33`:

```python
func = adapter.kabsch_umeyama if algo == "umeyama" else adapter.kabsch
```

Every other test that parametrises over `algo` uses `adapter.get_transform_func(algo)`. This one-off dispatch is fine for two algorithms but it will silently miss a new algorithm added to `get_transform_func()`. Prefer `adapter.get_transform_func(algo)` for consistency.

#### [NITPICK] `pytest_collection_modifyitems` does not filter Hypothesis tests by dim

`conftest.py:186-204` -- the hook filters out `(adapter, dim)` pairs only when both appear as pytest `callspec.params`. Hypothesis tests parametrised by `adapter` but with dim drawn inside `@given` are not filtered; they guard themselves with `assume(adapter.supports_dim(dim))` instead. This is correct and intentional, but a comment on the hook explaining why Hypothesis tests are excluded from its scope would clarify the design for future contributors.

### Verdict

APPROVED -- the architecturally significant fixture fix is correct, the adapter pattern is consistently extended, and the `strategies.py` module is correctly placed with no framework-specific logic. The TF bfloat16 gap is worth closing before it becomes a latent bug.

hunter-heidenreich:
## Test Quality Review

This PR adds substantial, meaningful test coverage with good naming and deterministic inputs throughout. The invariant tests in `test_properties.py`, degeneracy coverage in `test_degeneracy.py`, and double-backward coverage in `TestDoubleBackwardNonPyTorch` are all high-quality additions. A few issues are worth addressing.

### Issues

#### [RECOMMENDED] Absolute RMSD tolerance of 0.1 in descent-direction check is too loose

`test_gradient_verification.py:312`:

```python
assert rmsd_step <= rmsd_orig + 0.1, (...)
```

The comment correctly documents why a loose tolerance is used, but `0.1` is unconditional on the scale of `rmsd_orig`. If `rmsd_orig = 0.01` (which is possible at near-degenerate inputs where P and Q are very close), then `rmsd_orig + 0.1` allows the RMSD to increase tenfold before the test fails. The test would pass even for a gradient that points sharply uphill.

A relative bound like `rmsd_step <= rmsd_orig * 1.5 + 1e-4` would still accommodate SafeSVD's imprecision at degeneracies while catching badly wrong gradients. Alternatively, a larger step size guard (`if rmsd_orig < 1e-3: return`) would prevent the check from running when the baseline RMSD is so small that absolute noise dominates.

#### [RECOMMENDED] `np.testing.assert_allclose` used inconsistently across `test_properties.py`

Framework-parametrized tests in `TestKabschRecoveryND` (`test_properties.py:243-274`) use `np.testing.assert_allclose` with adapter tolerances:

```python
np.testing.assert_allclose(R, R_true, atol=adapter.atol * 100)
```

Every other framework-parametrized test in the suite uses `pytest.approx` with `rel=adapter.rtol, abs=adapter.atol`. For numpy-only tests the distinction is minor (both produce tolerance-based failures), but `pytest.approx` integrates better with pytest's failure display and diff formatting. The inconsistency is minor but makes `TestKabschRecoveryND` look like it belongs to a different convention.

#### [NITPICK] AAA structure is absent in several short `test_properties.py` tests

Tests like `test_rmsd_equals_transform_residual` (line 316) and `test_kabsch_rmsd_is_symmetric` (line 325) interleave act and assert without blank-line separation:

```python
R, t, rmsd = kabsch_np.kabsch(P_np, Q_np)
residual = np.linalg.norm(P_np @ R.T + t - Q_np) / np.sqrt(n)
np.testing.assert_allclose(float(rmsd), residual, atol=1e-8)
```

A single blank line before the assertion would clarify where the computation ends and the check begins, consistent with the other test files.

#### [NITPICK] `test_no_rotation_achieves_lower_rmsd` name is ambiguous

`test_properties.py:204` -- the test name reads as "no rotation achieves lower RMSD" (which is trivially true for large perturbations) rather than the intended "the returned rotation is locally optimal". Consider `test_returned_rotation_is_locally_optimal_kabsch` to make the claim self-documenting.

### Verdict

APPROVED -- the test coverage is meaningful, names are descriptive, random seeds are used consistently, and the new invariants test real properties that could catch real bugs. The loose RMSD descent tolerance is the only functionally significant gap.

hunter-heidenreich:
## Correctness Review

This PR contains no changes to `src/kabsch_horn/` source files -- it is a pure test-suite expansion. The review therefore focuses on whether the mathematical claims encoded in the new tests are correct and whether the test infrastructure (fixture rotations, RMSD formula, strategy geometry) correctly reflects the algorithms under test.

### No Issues Found

Every mathematical claim is correctly encoded:

**RMSD formula** (`test_properties.py:321`): `np.linalg.norm(P @ R.T + t - Q) / sqrt(n)` is the Frobenius norm of the residual matrix divided by sqrt(n), which equals sqrt(mean squared per-point distances) -- the correct RMSD definition.

**Rotation generation** (`conftest.py:14-24`, `strategies.py:38-48`): Both `_get_random_rotation` and `aligned_pair_nd/3d` generate proper rotations (det = +1) via QR decomposition with a column-flip correction when det < 0. The `conftest.py` variant applies an additional sign-diagonal trick (`Q * ph`) before the det check -- this is equivalent to `Q @ diag(sign(diag(R)))`, which is still orthogonal. Both approaches are correct.

**Invariant assertions**:
- RMSD symmetry (kabsch(P,Q).rmsd == kabsch(Q,P).rmsd): mathematically true since both solve the same minimax problem.
- Translation invariance of R: correct -- Kabsch centers both clouds, so translation cancels in the cross-covariance H.
- Uniform scale invariance of R: correct -- scaling both P and Q by c scales H by c^2, leaving U and V (and therefore R) unchanged.
- Rigid-transform invariance of RMSD: correct -- simultaneous rigid transform of both clouds preserves all inter-cloud distances.

**Umeyama scale recovery** (`test_properties.py:381-389`): For Q = c_true * (P @ R.T) + t, Umeyama must recover c_true. The `assume(sv[-1] > 1e-3)` guard prevents degenerate inputs where scale is unrecoverable. Tolerance `rtol=1e-4, atol=1e-4` is appropriate for float64 and exact inputs. Correct.

**`compute_sequential_expected_tensors`** (`utils.py:146-182`): The `np.ndindex(*batch_shape)` generalization is correct. `P_np[idx]` for a multi-index `idx` correctly slices a single unbatched cloud from an arbitrarily shaped batch. The shape reconstruction `np.empty(batch_shape + sample.shape)` handles scalar outputs (rmsd: shape `()`) correctly -- `batch_shape + ()` = `batch_shape`.

**xfail on JAX SVD double backward** (`test_gradient_verification.py:489-496`): `strict=True` is correct here -- if JAX upstream adds SVD JVP support, the test will catch the change and prompt removal of the mark.

**Near-collinear and near-coplanar strategies**: `nearly_collinear_3d` and `nearly_coplanar_nd` correctly model rank-deficient configurations. The `dim >= 3` guard in `nearly_coplanar_nd` is correct (coplanarity is undefined for 2D). The `assume(norm > 1e-10)` and `assume(range > 1.0)` guards in `nearly_collinear_3d` are appropriate.

### Verdict

APPROVED -- all mathematical properties are correctly stated and the supporting infrastructure (rotation generators, RMSD formula, batch indexing) is numerically and algorithmically sound.

PR #101, opened 2026-03-09
ci: pin action SHAs, add coverage artifacts, fix fail-fast and setup-python

### LINKED ISSUE #5: ci: workflow actions pinned to mutable tags, not commit SHAs
## Problem

All actions in `test.yml` are referenced by mutable version tags:

```yaml
uses: actions/checkout@v4
uses: astral-sh/setup-uv@v5
uses: actions/setup-python@v5
```

Mutable tags can be silently updated to point to different (potentially malicious) commits. Pinning to a full commit SHA is the supply-chain-hardened practice recommended by GitHub and OpenSSF.

## Impact

- A compromised action tag could execute arbitrary code in CI with access to repo secrets.
- Low likelihood but high severity if it occurs.

## Fix

Replace each tag reference with its corresponding commit SHA and keep the tag as a comment for readability. Example:

```yaml
# actions/checkout@v4
uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```

### Sub-tasks

- [ ] Resolve current SHAs for `actions/checkout@v4`, `astral-sh/setup-uv@v5`, and `actions/setup-python@v5`.
- [ ] Update `test.yml` with pinned SHAs.
- [ ] Consider adopting [Dependabot for GitHub Actions](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) to automate SHA bumps going forward.

hunter-heidenreich:
## Summary

- Pin all GitHub Actions refs to commit SHAs (inline tag comments for readability)
- Add `.github/dependabot.yml` for weekly Actions auto-update PRs
- Add `pytest-cov` to dev deps; pytest now emits `pytest-results.xml` and `coverage.xml`, uploaded as artifacts on every run (including failures)
- Set `fail-fast: false` on both `test-linux` and `test-macos` matrix jobs so a single failure doesn't cancel sibling legs
- Remove redundant `actions/setup-python` step; pass `python-version` directly to `setup-uv` instead

## Closes

Closes #5, #6, #7, #8

## Test plan

- [ ] CI runs on this PR -- confirm all matrix legs complete independently
- [ ] No `Set up Python` step appears in the Actions run log
- [ ] Artifacts section shows `test-results-linux-3.10` / `3.11` / `3.12` and `test-results-macos-*` containing `pytest-results.xml` and `coverage.xml`
- [ ] Workflow YAML references SHAs, not tags

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/pr-title-lint.yml
+++ b/.github/workflows/pr-title-lint.yml
@@ -15,6 +15,6 @@ jobs:
   lint-pr-title:
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5
+      - uses: amannn/action-semantic-pull-request@e32d7e603df1aa1ba07e981f2a23455dee596825  # v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -15,7 +15,7 @@ jobs:
     outputs:
       release_created: ${{ steps.release.outputs.release_created }}
     steps:
-      - uses: google-github-actions/release-please-action@v4
+      - uses: google-github-actions/release-please-action@e4dc86ba9405554aeba3c6bb2d169500e7d3b4ee  # v4
         id: release
         with:
           release-type: python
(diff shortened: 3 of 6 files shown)


hunter-heidenreich:
## pytest Review

The pytest-cov integration is correct and the flags are valid. One recommended improvement and one nitpick.

### Issues

#### [RECOMMENDED] Move `--cov` flags to `addopts` so local runs also produce coverage

`pyproject.toml` already has a `[tool.pytest.ini_options]` section:

```toml
[tool.pytest.ini_options]
testpaths = ["tests"]
```

The `--cov` and `--cov-report` flags are currently only present in the CI workflow command, so developers running `uv run pytest` locally get no coverage output. The idiomatic fix is to move those flags into `addopts`:

```toml
[tool.pytest.ini_options]
testpaths = ["tests"]
addopts = "--cov=src/kabsch_horn --cov-report=xml:coverage.xml"
```

The CI command then simplifies to:

```
uv run pytest --junit-xml=pytest-results.xml
```

`--junit-xml` stays in the workflow because it is CI-only output; coverage becomes a default for all runs.

#### [NITPICK] `tests/` on the CLI is redundant

`testpaths = ["tests"]` in `pyproject.toml` already tells pytest where to look. Passing `tests/` explicitly in the workflow command is harmless but redundant and would be removed automatically if coverage moves to `addopts`.

### Verdict

APPROVED -- the current implementation works correctly. The `addopts` change is recommended for developer experience but is not a blocker.

PR #103, opened 2026-03-09
feat(mlx): warn when float64 silently falls back to CPU

### LINKED ISSUE #51: feat(mlx): warn when float64 silently falls back to CPU
## Problem

`MLXAdapter._set_device()` calls `mx.set_default_device(mx.cpu)` for float64 without any user-visible indication. Apple Silicon GPUs do not support true float64, so CPU fallback is correct, but it is silent.

A user benchmarking MLX float64 and expecting GPU execution would receive CPU timings with no warning, which could produce misleading benchmark comparisons (see related benchmark issues #20-#25).

The same pattern likely applies to the production code in `src/kabsch_horn/mlx/` if users ever call it directly with float64 tensors.

## Fix

Emit a `warnings.warn` when float64 is detected and device is forced to CPU:

```python
import warnings

def _set_device(self) -> None:
    if self.precision == "float64":
        warnings.warn(
            "MLX does not support float64 on GPU; falling back to CPU.",
            UserWarning,
            stacklevel=2,
        )
        mx.set_default_device(mx.cpu)
    else:
        mx.set_default_device(mx.gpu)
```

hunter-heidenreich:
## Summary

- Adds `_warn_if_float64` helper to both MLX modules (`kabsch_svd_nd.py`, `horn_quat_3d.py`) that emits a `UserWarning` and calls `mx.set_default_device(mx.cpu)` when float64 input is detected
- All four public functions (`kabsch`, `kabsch_umeyama`, `horn`, `horn_with_scale`) call the helper at entry, so rmsd wrappers inherit it transitively
- `MLXAdapter._set_device` in `tests/adapters.py` gains the same `warnings.warn` so test infrastructure is consistent
- New test file `tests/test_mlx_float64_warning.py` with 12 tests covering: warning fires for all 6 public functions, no warning for float32, adapter warns on float64, adapter silent on float32

## Test plan

- [ ] `uv run pytest tests/test_mlx_float64_warning.py -v` -- all 12 pass
- [ ] `uv run pytest tests/ -q` -- no regressions
- [ ] `uv run ruff check . && uv run ruff format --check .` -- clean

Closes #51

### DIFF
--- a/src/kabsch_horn/mlx/_utils.py
+++ b/src/kabsch_horn/mlx/_utils.py
@@ -1,0 +1,27 @@
+import warnings
+
+import mlx.core as mx
+
+
+def _warn_if_float64(P: mx.array, Q: mx.array, stacklevel: int = 3) -> None:
+    """Emit a UserWarning if either P or Q is float64.
+
+    MLX does not support float64 on GPU. Passing float64 inputs sets the
+    process-wide default device to CPU (mx.set_default_device(mx.cpu)).
+    This is a persistent side effect: all subsequent MLX operations in the
+    process will also run on CPU until the default device is changed again.
+    stacklevel=3 is correct for direct callers (kabsch, horn, etc.).
+    For rmsd wrappers (kabsch_rmsd -> kabsch -> here), stacklevel points
+    one level into the wrapper rather than the user call site; this is an
+    accepted limitation.
+    """
+    if P.dtype == mx.float64 or Q.dtype == mx.float64:
+        warnings.warn(
+            "MLX does not support float64 on GPU; falling back to CPU. "
+            "This sets the process-wide default device to CPU "
+            "(mx.set_default_device(mx.cpu)) and will affect all subsequent "
+            "MLX operations in this process.",
+            UserWarning,
+            stacklevel=stacklevel,
+        )
+        mx.set_default_device(mx.cpu)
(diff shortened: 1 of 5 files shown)


hunter-heidenreich:
## Test Architecture Review

The new test file is well-isolated and correctly avoids the adapter layer for warning-behavior tests. One structural issue in the adapter is worth addressing: `_set_device` now emits a warning that duplicates the library warning on every method dispatch.

### Issues

#### [RECOMMENDED] `MLXAdapter._set_device` warning fires twice per adapter call for float64

`tests/adapters.py` lines 396-407: `_set_device()` is called at the top of every adapter method (`convert_in`, `kabsch`, `kabsch_umeyama`, `horn`, `horn_with_scale`, `kabsch_rmsd`, `kabsch_umeyama_rmsd`, `is_nan`, `get_grad`). For float64, each of those calls now emits a `UserWarning`. Then the delegated library function (`kabsch_mlx.kabsch`, etc.) calls `_warn_if_float64(P, Q)` and emits a second warning.

A single typical test invocation -- `convert_in` (warning 1) + `adapter.kabsch` dispatch (warning 2) + library `_warn_if_float64` (warning 3) -- produces three warnings. The full suite output already shows 1166 warnings. This will grow with every new test that uses the float64 MLX adapter.

The library now owns the user-facing warning. The adapter's role is to set the correct device before converting inputs, not to warn. The cleanest fix is to drop `warnings.warn` from `_set_device` entirely and rely on the library functions to emit the warning. `test_mlx_adapter_float64_emits_warning` can then be removed or replaced with a test that calls a library function via the adapter and checks the library warning fires.

### Verdict

APPROVED

PR #104, opened 2026-03-09
test(horn): add P!=Q near-collinear gradient stability test

### LINKED ISSUE #92: test: near-collinear gradient tests only exercise P==Q, missing P!=Q variant
## Summary

`test_gradients_stable_nearly_collinear_hypothesis` in `tests/test_differentiability_traps.py` always sets `Q_np = P_np.copy()`, so it only ever tests the P==Q singularity (zero displacement / RMSD=0).

The realistic and more interesting failure mode -- two *different* near-collinear clouds where the cross-covariance matrix is both ill-conditioned *and* non-zero -- is never exercised.

## Why it matters

The `SafeSVD`/`SafeEigh` gradient wrappers are specifically designed to handle ill-conditioned input. Testing only the zero-displacement case gives false confidence: a regression that broke gradient stability for `P != Q` near-collinear inputs would pass undetected.

## Suggested fix

Add a second variant of the test where `Q_np` is a small rigid-body perturbation of `P_np` (e.g. apply a rotation within the collinear plane + small translation), so the cloud is still near-collinear but `rmsd > 0`.

## Relevant files

- `tests/test_differentiability_traps.py` (~lines 304, 397)
- `tests/strategies.py` -- `nearly_collinear_3d` strategy

hunter-heidenreich:
## Summary

- The existing `test_gradients_stable_nearly_collinear_hypothesis` in `TestHornDifferentiabilityTraps` always set `Q = P.copy()`, only ever testing the zero-displacement singularity
- Adds `test_gradients_stable_nearly_collinear_different_clouds` to the Horn class, drawing P and Q independently via `@given(nearly_collinear_3d(), nearly_collinear_3d())`
- Mirrors the equivalent test that already exists in `TestKabschDifferentiabilityTraps`
- float16/bfloat16 skipped (gradients overflow), float32/float64 pass across all frameworks

## Test plan

- [ ] `uv run pytest tests/test_differentiability_traps.py -k "nearly_collinear_different_clouds and Horn" -v` -- 32 passed, 32 skipped
- [ ] `uv run ruff check . && uv run ruff format --check .` -- clean

Closes #92

### DIFF
--- a/tests/test_differentiability_traps.py
+++ b/tests/test_differentiability_traps.py
@@ -616,6 +616,45 @@ class TestHornDifferentiabilityTraps:
     @pytest.mark.parametrize("adapter", frameworks)
     @pytest.mark.parametrize("wrt", ["P", "Q"])
     @pytest.mark.parametrize("algo", ["horn", "horn_with_scale"])
+    def test_gradients_stable_nearly_collinear_different_clouds(
+        self,
+        adapter: FrameworkAdapter,
+        wrt: str,
+        algo: str,
+    ) -> None:
+        """Horn gradients remain finite when P != Q are near-collinear (Hypothesis).
+
+        Compared to test_gradients_stable_nearly_collinear_hypothesis (which sets
+        Q = P), this test draws P and Q independently so the cross-covariance H
+        is rank-1 but non-zero -- the realistic hard case SafeEigh is designed for.
+        """
+        if getattr(adapter, "precision", "float64") in ("float16", "bfloat16"):
+            pytest.skip(
+                "Near-collinear independent clouds produce large eigh gradients "
+                "that overflow float16 range."
+            )
+
+        @settings(
+            max_examples=30,
+            suppress_health_check=[HealthCheck.too_slow, HealthCheck.filter_too_much],
+            deadline=None,
+        )
+        @given(nearly_collinear_3d(), nearly_collinear_3d())
+        def _inner(P_np: np.ndarray, Q_np: np.ndarray) -> None:
+            n = min(P_np.shape[0], Q_np.shape[0])
+            P_np_t = P_np[:n]
+            Q_np_t = Q_np[:n]
+            P = adapter.convert_in(P_np_t.astype(np.float64))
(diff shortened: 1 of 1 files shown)


hunter-heidenreich:
## pytest Review

The new test is structurally identical to the Kabsch counterpart (`test_gradients_stable_nearly_collinear_different_clouds` in `TestDifferentiabilityTraps`) and all other Hypothesis-wrapped tests in the file. Every pytest idiom -- the `_inner` closure, runtime `pytest.skip()` for precision checks, parametrize ordering, and absence of `pytest.param` wrappers -- is consistent with the established conventions.

No issues found.

### Verdict

APPROVED

hunter-heidenreich:
## Test Quality Review

The new test fills a genuine coverage gap: the existing `test_gradients_stable_nearly_collinear_hypothesis` only tested P=Q (zero-displacement, trivial gradient), while the realistic SafeEigh stress case is P≠Q with a rank-1 non-zero H. The addition is meaningful and the assertion (`np.all(np.isfinite(grad))`) is the right property to check.

### Issues

#### [NITPICK] Skip condition is broader than the sibling test, with no inline explanation

`test_gradients_stable_nearly_collinear_hypothesis` (the P=Q sibling at line 591) skips only `horn_with_scale` + float16/bfloat16, because the overflow there comes from variance division. The new test skips **all algos** (including base `horn`) for float16/bfloat16. The skip message explains the overflow ("large eigh gradients") but does not explain why base `horn` -- which the sibling runs fine at float16 when P=Q -- now also overflows.

A one-line comment like:
```python
# P != Q makes H non-zero, amplifying eigh gradients enough to overflow float16
# even for base horn (unlike the P=Q case which is near-zero gradient).
```
would make this self-contained for anyone comparing the two tests. The Kabsch counterpart (line 333) has the same gap, so this is a pre-existing pattern, not unique to this PR.

### Verdict

APPROVED

hunter-heidenreich:
## Code Style Review

The single changed file passes `ruff check` cleanly, all writing conventions are followed, and type annotations are correct.

No issues found.

### Verdict

APPROVED

PR #106, opened 2026-03-09
fix: autograd safety, dtype-aware eps, smooth RMSD gradients, cross-framework cleanup, docs site

### LINKED ISSUE #5: ci: workflow actions pinned to mutable tags, not commit SHAs

### LINKED ISSUE #6: ci: no test results artifact or coverage reporting
## Problem

Pytest runs but produces no machine-readable output. There is no JUnit XML artifact, no coverage measurement, and no coverage badge. This makes it hard to track which tests failed in a given run without reading raw logs, and there is no long-term visibility into coverage trends.

## Impact

- No per-test failure visibility in the GitHub Actions UI.
- No coverage badge for the README.
- Regressions in coverage are invisible.

### Sub-tasks

- [ ] Add `--junit-xml=pytest-results.xml` to the pytest invocation and upload it with `actions/upload-artifact` so GitHub can surface per-test results.
- [ ] Add `pytest-cov` to the dev dependencies and run with `--cov=src/kabsch_horn --cov-report=xml`.
- [ ] Integrate with a coverage service (e.g. Codecov or Coveralls) and add the badge to `README.md`.

### LINKED ISSUE #7: ci: matrix fail-fast cancels sibling runs on first failure
## Problem

GitHub Actions defaults `fail-fast: true` on matrix jobs. When the Python 3.10 leg fails, the 3.11 and 3.12 legs are cancelled. For a library that explicitly targets all three versions, this hides whether a failure is version-specific or universal.

## Impact

- A 3.10-only regression looks the same as a universal failure.
- Debugging requires a re-run just to see results on other versions.

## Fix

Add `fail-fast: false` to the matrix strategy in `test.yml`:

```yaml
strategy:
  fail-fast: false
  matrix:
    python-version: ["3.10", "3.11", "3.12"]
```

### LINKED ISSUE #8: ci: redundant actions/setup-python step alongside setup-uv
## Problem

The workflow runs both `astral-sh/setup-uv` and `actions/setup-python`. When using uv, Python version management can be handled directly by `setup-uv` via its `python-version` input, making the `setup-python` step redundant.

## Impact

- Minor overhead on every run.
- Two competing Python version managers can cause subtle conflicts.

## Fix

Remove the `actions/setup-python` step and pass the Python version to `setup-uv` instead:

```yaml
- name: Install uv
  uses: astral-sh/setup-uv@v5
  with:
    enable-cache: true
    python-version: ${{ matrix.python-version }}
```

### LINKED ISSUE #9: ci: CI/CD overhaul tracking issue
Tracking issue for all identified CI/CD gaps. Each sub-issue is scoped and actionable on its own.

## Critical

- [ ] #2 CI does not run on the dev branch
- [ ] #1 No CD pipeline -- releases are fully manual

## Severe

- [ ] #3 MLX cannot install on ubuntu-latest runners
- [ ] #4 No dependency caching -- cold installs on every run

## Moderate

- [ ] #5 Matrix fail-fast cancels sibling runs on first failure

## Low

- [ ] #6 Redundant actions/setup-python step alongside setup-uv
- [ ] #7 No test results artifact or coverage reporting
- [ ] #8 Workflow actions pinned to mutable tags, not commit SHAs

### LINKED ISSUE #27: docs: investigate and document torch.compile compatibility
## Goal

Determine whether `SafeSVD` and `SafeEigh` (PyTorch custom autograd functions) are compatible with `torch.compile`, document the findings, and fix any incompatibilities found.

## Background

Both backward passes use in-place mutations on views:

```python
# pytorch/kabsch_svd_nd.py:64,68
safe_D.diagonal(dim1=-2, dim2=-1).fill_(1.0)
F.diagonal(dim1=-2, dim2=-1).zero_()

# pytorch/horn_quat_3d.py (SafeEigh.backward -- same pattern)
safe_D.diagonal(dim1=-2, dim2=-1).fill_(1.0)
F.diagonal(dim1=-2, dim2=-1).zero_()
```

In-place mutations on views (`.diagonal()` returns a view) inside `autograd.Function.backward` are legal in eager mode but can confuse the `torch.compile` graph tracer, which tracks aliasing conservatively. There may also be issues with `torch.compile`'s handling of custom `Function` classes in general.

## Investigation Steps

1. Run `torch.compile(kabsch)(P, Q)` and `torch.compile(kabsch_rmsd)(P, Q)` -- do they execute without error?
2. Run with `torch.compile(..., fullgraph=True)` -- does graph capture succeed or fall back to eager?
3. Trigger the backward pass through a compiled `kabsch_rmsd` -- any graph breaks or errors?
4. Repeat for `horn` / `horn_with_scale`
5. Test with `backend="inductor"` (default) and `backend="eager"` for isolation

## Expected Deliverables

- A clear compatibility statement: "works / works with caveats / broken" for `torch.compile`
- If broken: a fix (likely replacing in-place `.fill_`/`.zero_` with `torch.where` or explicit construction)
- If working: a test added to the test suite that compiles `kabsch_rmsd` and runs a forward+backward pass
- A note in the README PyTorch section on `torch.compile` support status

### LINKED ISSUE #28: docs: JAX jit usage guidance and performance recommendation
## Goal

Document that JAX users should wrap calls in `jax.jit` for performance, and quantify the cost of not doing so (see companion profiling issue #25).

## Background

The JAX functions (`kabsch`, `kabsch_umeyama`, `horn`, `horn_with_scale`, `kabsch_rmsd`) are not decorated with `@jax.jit`. Without JIT, JAX traces the Python function on every call, incurring overhead that can be significant in a training loop. The custom VJP (`@jax.custom_vjp` on `safe_svd`) compounds this slightly.

This is standard JAX behavior and not a bug, but the library gives users no signal that they should JIT-wrap.

## Proposed Changes

1. **Docstring update**: Add a "Performance" note to each public JAX function:
   > For best performance, wrap this function with `jax.jit`. Calling without JIT incurs tracing overhead on every call.

2. **README update**: Add a JAX usage example showing the recommended pattern:
   ```python
   import jax
   from kabsch_horn.jax import kabsch_rmsd

   # Recommended: JIT-compile once
   fast_rmsd = jax.jit(kabsch_rmsd)
   loss = fast_rmsd(P, Q)
   ```

3. **(Optional) convenience exports**: Consider whether the JAX `__init__.py` should export pre-jitted variants (e.g. `kabsch_jit`, `kabsch_rmsd_jit`). This is a minor API question -- it avoids user footguns but adds surface area.

## Acceptance Criteria

- Docstrings updated on all public JAX functions
- README JAX section includes a JIT usage example
- Decision recorded on whether to export pre-jitted variants

### LINKED ISSUE #30: perf: replace repeated ones-stack pattern with single-allocation in all frameworks
## Goal

Replace the `torch.stack([ones] * (D-1) + [d_sign], dim=-1)` pattern (and its equivalents in JAX, TF, NumPy, MLX) with a single-allocation approach.

## Background

In every `kabsch` and `kabsch_umeyama` implementation, the reflection correction diagonal is built by stacking D-1 identical `ones` tensors with a sign vector:

```python
# pytorch/kabsch_svd_nd.py:178
B_diag = torch.stack([ones] * (D - 1) + [d_sign], dim=-1)  # BxD

# jax/kabsch_svd_nd.py:137
B_diag = jnp.stack([ones] * (D - 1) + [d_sign], axis=-1)
```

This creates D-1 redundant `ones` tensors in Python, each needing a separate allocation and a stack op. For D=3 it's only 2 allocations -- cheap in absolute terms -- but it's semantically wasteful and the pattern repeats across 10+ call sites in the codebase.

The cleaner approach is a single buffer fill + index assignment:

```python
# PyTorch equivalent
B_diag = torch.ones(B, D, dtype=P.dtype, device=P.device)
B_diag[:, -1] = d_sign
```

This is one allocation, one scalar write, no Python list construction.

## Scope

All occurrences across:
- `pytorch/kabsch_svd_nd.py` (2 occurrences)
- `jax/kabsch_svd_nd.py` (2 occurrences)
- `tensorflow/kabsch_svd_nd.py` (2 occurrences)
- `numpy/kabsch_svd_nd.py` (2 occurrences)
- `mlx/kabsch_svd_nd.py` -- different pattern already (uses `mx.concatenate`), assess separately

## Acceptance Criteria

- All occurrences replaced with the single-allocation pattern
- Existing tests pass unchanged (behavior is identical)
- No regressions in gradient tests (the `B_diag` participates in the autograd graph)

### LINKED ISSUE #31: fix: replace runtime vmap_diag branch in JAX SVD backward with diag_embed
## Goal

Remove the Python-level conditional in the JAX SVD VJP backward that selects between `jnp.diag` and `jax.vmap(jnp.diag)` based on tensor rank at trace time.

## Background

In `jax/kabsch_svd_nd.py`, the VJP backward contains:

```python
# jax/kabsch_svd_nd.py:74-76
vmap_diag = jax.vmap(jnp.diag) if S.ndim > 1 else jnp.diag
S_diag = vmap_diag(grad_S)
S_mat = vmap_diag(S)
```

This is a Python conditional on `S.ndim`, which is evaluated at JAX trace time (inside `jax.jit`). The two branches produce structurally different JAX functions, meaning a JIT-compiled function compiled for batched input cannot be reused for unbatched input (and vice versa) -- they trace to different graphs. This is generally harmless (JAX retraces for new shapes) but it's unnecessary complexity.

The `jax.vmap(jnp.diag)` pattern is exactly what `jnp.diag_embed` is for -- it generalizes `diag` to batched inputs:

```python
# Replacement
S_diag = jnp.diag_embed(grad_S)  # works for any leading batch dims
S_mat = jnp.diag_embed(S)
```

(`jnp.diag_embed` requires JAX >= 0.4.1, which is already a transitive dependency.)

## Fix

Replace both occurrences in `_bwd`:

```python
# Before
vmap_diag = jax.vmap(jnp.diag) if S.ndim > 1 else jnp.diag
S_diag = vmap_diag(grad_S)
S_mat = vmap_diag(S)

# After
S_diag = jnp.diag_embed(grad_S)
S_mat = jnp.diag_embed(S)
```

## Acceptance Criteria

- `vmap_diag` conditional removed
- All existing JAX tests pass
- Gradient correctness tests pass (the diag is in the hot path of the backward)
- No JAX version compatibility regression (verify `jnp.diag_embed` availability)

### LINKED ISSUE #51: feat(mlx): warn when float64 silently falls back to CPU

### LINKED ISSUE #90: docs: document cross-covariance conditioning requirement for rotation invariance
## Summary

The rotation returned by Kabsch/Umeyama is only invariant to global translation and uniform scaling when the cross-covariance matrix `H = P_c.T @ Q_c` is well-conditioned. When the smallest singular value of H is near zero, a small perturbation (rigid shift or scaling of both P and Q) can tip the SVD into a different branch, producing a different rotation.

This was discovered when writing `TestAlignmentInvariants` in `tests/test_properties.py`. Tests `test_r_invariant_to_translation` and `test_r_invariant_to_uniform_scale` require `assume(sv_H[-1] > 0.1)` to pass reliably.

## Mathematical context

The SVD of H = U S V^T is used to construct R = V U^T. When two singular values of H are equal (or near-equal), U and V are not unique -- any rotation in the corresponding invariant subspace is a valid solution. A tiny perturbation can select a different element of that set, making R numerically unstable under translation/scale.

## Recommended action

- Add a note to the public API docstrings explaining this conditioning requirement
- Consider adding a `condition_number` or `rank` diagnostic to the return value (or a separate utility) so users can detect degenerate configurations at call time
- Alternatively, document the `eps` parameter in `SafeSVD`/`SafeEigh` as the relevant threshold

## Relevant test

`tests/test_properties.py::TestAlignmentInvariants::test_r_invariant_to_translation` and `test_r_invariant_to_uniform_scale` (require `assume(sv_H[-1] > 0.1)`)

### LINKED ISSUE #92: test: near-collinear gradient tests only exercise P==Q, missing P!=Q variant

### LINKED ISSUE #93: test: nearly_coplanar_nd strategy produces collinear clouds for dim=2
## Summary

The `nearly_coplanar_nd` Hypothesis strategy in `tests/strategies.py` zeroes out `P[:, -1]` to force near-coplanarity. When `dim=2`, this zeroes `P[:, 1]`, collapsing all points onto the x-axis -- producing a *collinear* cloud, not a coplanar one.

There is also no meaningful notion of "nearly coplanar" in 2D (the whole 2D plane is the space), so the strategy is semantically undefined for `dim=2`.

## Why it matters

Tests consuming this strategy check `adapter.supports_dim(dim)` but do not guard against the `dim=2` semantic mismatch. For `dim=2` inputs the strategy silently produces a different degeneracy (collinear) than intended (coplanar), which could mask or misattribute failures.

## Suggested fix

Add `assume(dim >= 3)` at the top of `nearly_coplanar_nd`, or change the strategy's `dim` draw to `st.integers(min_value=3, max_value=...)` directly, and document that `dim=2` is excluded.

## Relevant files

- `tests/strategies.py` (~lines 93-106, `nearly_coplanar_nd`)

### LINKED ISSUE #110: refactor: replace hardcoded eps=1e-12 with dtype-aware floor
## Summary

Several places across all frameworks use a hardcoded `1e-12` as a numerical floor. This works, but is not tied to the actual precision of the working dtype. Replacing it with a dtype-aware value would be more principled and self-documenting.

## Locations

The `1e-12` appears in at least three roles per framework (PyTorch, JAX, TensorFlow, MLX):

1. **RMSD sqrt floor** -- `clamp(sum_sq / N, min=1e-12)` before `sqrt`, preventing `d/dx sqrt(0) = inf`
2. **Scale denominator** -- `clamp(var_P, min=1e-12)` preventing division by zero in `horn_with_scale` / `kabsch_umeyama`
3. **SafeEigh / SafeSVD backward eps** -- masks near-zero eigenvalue/singular-value differences in the custom autograd backward

## Proposed change

All frameworks upcast float16/bfloat16 to float32 internally, so the working dtype is always float32 or float64. A dtype-aware floor would use something like:

- `torch.finfo(working_dtype).eps` (float32: ~1.19e-7) -- gentler gradient bound, slightly larger forward error
- `torch.finfo(working_dtype).eps ** 2` (float32: ~1.4e-14) -- smaller forward error, larger gradient bound

Items 1 and 2 are straightforward to make dtype-aware.

## Caution on item 3 (SafeEigh / SafeSVD)

The backward eps for SafeEigh/SafeSVD controls what eigenvalue spacing counts as "degenerate." Making this too small could reintroduce NaN gradients for symmetric or near-symmetric point clouds. This item warrants separate analysis before changing.

## Why now

The `1e-12` floor in the RMSD sqrt was recently discussed in #108. Tracking this as a follow-up keeps the PR focused while capturing the broader clean-up opportunity.

### LINKED ISSUE #111: fix(pytorch): in-place view mutations in SafeSVD/SafeEigh backward block higher-order gradients and torch.compile
## Problem

Both `SafeSVD.backward` and `SafeEigh.backward` use `.diagonal(dim1=-2, dim2=-1).fill_(1.0)` and `.diagonal(dim1=-2, dim2=-1).zero_()` to set diagonal values. `.diagonal()` returns a view, so `.fill_()` and `.zero_()` are in-place mutations through a view.

**Locations:**
- `src/kabsch_horn/pytorch/kabsch_svd_nd.py:73-77`
- `src/kabsch_horn/pytorch/horn_quat_3d.py:33-35`

## Impact

1. **Higher-order gradients**: If a user calls with `create_graph=True` (for Hessian-vector products, meta-learning, etc.), these in-place mutations will raise a `RuntimeError` or silently produce incorrect higher-order gradients, because PyTorch's autograd cannot track in-place ops on graph intermediates when building a higher-order graph.

2. **torch.compile**: These patterns are a known source of graph breaks under `torch.compile` (Dynamo + Inductor). In-place mutation through views forces the compiler to insert a graph break and fall back to eager execution.

## Suggested Fix

Replace in-place view mutations with out-of-place operations:

```python
eye = torch.eye(safe_D.shape[-1], dtype=safe_D.dtype, device=safe_D.device)
safe_D = torch.where(eye.bool(), torch.ones_like(safe_D), safe_D)
F = 1.0 / safe_D
F = torch.where(eye.bool(), torch.zeros_like(F), F)
```

Found by automated audit.

### LINKED ISSUE #112: docs(pytorch): SVD backward comment contradicts code's sign convention
## Problem

The comment on lines 43-45 of `src/kabsch_horn/pytorch/kabsch_svd_nd.py` states the formula as `dA = U (diag(dS) + J S + S K) V^T` with positive signs. The actual code on line 92 computes `term = S_diag - J @ S_mat - S_mat @ K` with negative signs.

The discrepancy arises because `J` and `K` are defined with an explicit antisymmetrization (`Ut_dU - Ut_dU.mH`) that introduces a sign/scaling difference compared to the standard formulation. The code is mathematically correct (passes finite-difference gradient verification), but the comment will mislead anyone trying to verify the implementation against the formula.

## Suggested Fix

Either update the comment to match the code's sign convention, or add a note explaining the relationship between the two.

Found by automated audit.

### LINKED ISSUE #113: docs(pytorch): higher-order gradients unsupported and undocumented in SafeSVD/SafeEigh
## Problem

`SafeSVD` and `SafeEigh` only support first-order backward. Passing `create_graph=True` (required for Hessian-vector products, higher-order optimization, or MAML-style meta-learning) will fail due to:

1. In-place view mutations in backward (not differentiable)
2. Discrete `torch.where` masks (produce zero gradients for the mask condition)

**Locations:**
- `src/kabsch_horn/pytorch/kabsch_svd_nd.py:4-11`
- `src/kabsch_horn/pytorch/horn_quat_3d.py:4-7`

Neither the class docstrings nor the public function docstrings mention this limitation.

## Suggested Fix

Add a note such as "Only first-order gradients are supported; higher-order differentiation via `create_graph=True` is not supported" to the class and/or public function docstrings.

Found by automated audit.

### LINKED ISSUE #114: refactor(pytorch): use torch.linalg.det instead of torch.det
## Problem

`src/kabsch_horn/pytorch/kabsch_svd_nd.py` lines 182 and 292 use `torch.det(...)`. While `torch.det` is not yet formally deprecated and delegates to `torch.linalg.det` internally, the file already uses `torch.linalg.svd` and the Horn file uses `torch.linalg.eigh`. Using `torch.linalg.det` would be consistent with the modern `torch.linalg` namespace used elsewhere. The PyTorch docs note that `torch.det` may be deprecated in a future release.

Found by automated audit.

### LINKED ISSUE #115: perf(pytorch): ctx.needs_input_grad not checked in SafeSVD/SafeEigh backward
## Problem

Neither `SafeSVD.backward` nor `SafeEigh.backward` checks `ctx.needs_input_grad[0]` before computing `grad_A`. When the input tensor does not require gradients, the entire backward computation (matrix multiplications, F-matrix construction, etc.) runs unnecessarily.

**Locations:**
- `src/kabsch_horn/pytorch/kabsch_svd_nd.py:37-96`
- `src/kabsch_horn/pytorch/horn_quat_3d.py:17-43`

## Suggested Fix

Gate the computation:
```python
if not ctx.needs_input_grad[0]:
    return None, None
```

In practice inputs almost always require gradients, so this is a minor efficiency concern, but it follows PyTorch best practices for `torch.autograd.Function`.

Found by automated audit.

### LINKED ISSUE #116: fix(pytorch): SafeSVD NaN-fallback saves wrong-shape tensor for Vh
## Problem

In the `LinAlgError` exception handler at `src/kabsch_horn/pytorch/kabsch_svd_nd.py:23-27`, `nan_mat` is created with shape `A.shape` (i.e., `[..., M, N]` for a general rectangular input) and is saved as both `U` and `Vh` via `ctx.save_for_backward(nan_mat, nan_s, nan_mat)`.

The backward then does `Vh.mH` (line 50) and matrix multiplications assuming `Vh` has shape `[..., D, D]`. Since inputs to `SafeSVD` are always square cross-covariance matrices in this codebase, the shape mismatch is not triggered in practice. However, if this function were ever used on non-square inputs, the NaN fallback would produce shape errors in backward.

This is a latent issue given the current usage pattern.

Found by automated audit.

### LINKED ISSUE #117: refactor(pytorch): safe_svd unbatching path is dead code
## Problem

`safe_svd` at `src/kabsch_horn/pytorch/kabsch_svd_nd.py:110-118` checks `A.ndim == 2` and adds a batch dimension, then squeezes it back. However, both internal callers (`kabsch` and `kabsch_umeyama`) already reshape their input `H` to be at least 3D before calling `safe_svd`. `safe_svd` is not exported in `__init__.py`, so no external caller can reach it.

The ndim==2 handling is unreachable dead code. Removing it would simplify the function.

Found by automated audit.

### LINKED ISSUE #119: refactor(jax): use isinstance for SymbolicZero check instead of type()
## Problem

The SymbolicZero checks in the JAX SVD and eigh backward passes use `type(grad_U) is jax.custom_derivatives.SymbolicZero` (strict identity) rather than `isinstance(grad_U, jax.custom_derivatives.SymbolicZero)`.

**Locations:**
- `src/kabsch_horn/jax/kabsch_svd_nd.py:36-50`
- `src/kabsch_horn/jax/horn_quat_3d.py:23-32`

`isinstance` is the standard Python pattern for type checks and would be resilient to future subclassing.

Found by automated audit.

### LINKED ISSUE #120: fix(mlx): duplicate shape-equality check in horn/horn_with_scale
## Problem

Both `horn` and `horn_with_scale` contain an identical `P.shape != Q.shape` check twice -- once before `_warn_if_float64` and once after. This is a copy-paste artifact; nothing between the two checks can change the shapes.

**Locations:**
- `src/kabsch_horn/mlx/horn_quat_3d.py:63-66` and `68-71`
- `src/kabsch_horn/mlx/horn_quat_3d.py:177-180` and `182-185`

The second check in each function is dead code and should be removed.

Found by automated audit.

### LINKED ISSUE #121: refactor(mlx): inconsistent mx.array() input wrapping between Kabsch and Horn
## Problem

`horn` and `horn_with_scale` wrap inputs with `P = mx.array(P); Q = mx.array(Q)`, which silently converts lists, numpy arrays, etc. to MLX arrays. The `kabsch` and `kabsch_umeyama` functions do not perform this conversion and would raise an `AttributeError` on non-MLX inputs.

**Locations:**
- `src/kabsch_horn/mlx/horn_quat_3d.py:61-62` and `175-176`
- `src/kabsch_horn/mlx/kabsch_svd_nd.py` (no conversion)

The API should be consistent: either all functions accept non-MLX inputs or none do.

Found by automated audit.

### LINKED ISSUE #122: refactor(mlx): unnecessary scalar mx.array construction for division
## Problem

`kabsch_umeyama` constructs `N = mx.array(P.shape[-2], dtype=P.dtype)` and `horn_with_scale` constructs `N_pts_f = mx.array(P.shape[-2], dtype=P.dtype)` solely to divide arrays. MLX supports dividing an array by a Python int or float directly, which avoids creating a scalar array node in the computation graph.

**Locations:**
- `src/kabsch_horn/mlx/kabsch_svd_nd.py:226`
- `src/kabsch_horn/mlx/horn_quat_3d.py:192`

Found by automated audit.

### LINKED ISSUE #123: fix(tensorflow): safe_svd backward eye-based masking produces inf/NaN for near-equal singular values
## Problem

The masking logic in the SVD backward at `src/kabsch_horn/tensorflow/kabsch_svd_nd.py:35`:

```python
tf.where(tf.abs(S_sq_diff) < 1e-12, eye * 1e-12, S_sq_diff)
```

uses the identity matrix to fill small values. For **off-diagonal** entries where singular values are nearly equal, `eye[i,j] = 0`, so the replacement is `0 * 1e-12 = 0`. This produces `F = 1/0 = inf` on those entries, which propagates `inf` or `NaN` through the gradient.

This defeats the entire purpose of the safe SVD wrapper. The `call_safe_eigh` backward in `horn_quat_3d.py:21-22` already handles this correctly with `eps * tf.sign(D + eps)`, as do the PyTorch and JAX SVD backwards.

## Suggested Fix

Replace the eye-based masking with `eps * tf.sign(S_sq_diff + eps)`, then separately set the diagonal of `safe_D` to `1.0` and zero the diagonal of `F`, matching the pattern in the other frameworks.

Found by automated audit.

### LINKED ISSUE #124: fix(tensorflow): tf.rank() in Python if-statement breaks @tf.function tracing
## Problem

`is_single = tf.rank(P) == 2` produces a `tf.Tensor`, not a Python bool. While this works in eager mode (auto-evaluation), wrapping `horn` or `horn_with_scale` in `@tf.function` will raise `OperatorNotAllowedInGraphError` because Python `if` on a symbolic tensor is not allowed during tracing.

**Locations:**
- `src/kabsch_horn/tensorflow/horn_quat_3d.py:55,132,159,242`

## Suggested Fix

Use `len(P.shape)` or `P.ndim`, both of which return Python ints for statically-shaped tensors. This matches PyTorch's `P.ndim == 2` pattern.

Found by automated audit.

### LINKED ISSUE #125: refactor(tensorflow): safe_eigh has unused eps parameter
## Problem

`safe_eigh(A, eps=1e-12)` at `src/kabsch_horn/tensorflow/horn_quat_3d.py:4-5` accepts an `eps` parameter but ignores it entirely, just returning `tf.linalg.eigh(A)`. The actual eps is hardcoded inside `call_safe_eigh`'"'"'s grad closure (line 11).

The function signature implies eps controls gradient safety, but it does nothing.

## Suggested Fix

Either remove the parameter or forward it into `call_safe_eigh`.

Found by automated audit.

### LINKED ISSUE #126: docs(tensorflow): call_safe_eigh/safe_eigh and horn functions lack docstrings and type annotations
## Problem

`safe_svd` in `kabsch_svd_nd.py` has type annotations and a docstring. The corresponding functions in `horn_quat_3d.py` have neither:

- `safe_eigh` (line 4) -- no annotations, no docstring
- `call_safe_eigh` (line 8-9) -- no annotations, no docstring
- `horn` (line 39) -- no docstring
- `horn_with_scale` (line 141) -- no docstring

For consistency with the SVD wrapper, other frameworks (JAX has annotations on its `safe_eigh`), and the TF `kabsch`/`kabsch_umeyama` functions which all have full docstrings.

Found by automated audit.

### LINKED ISSUE #127: fix(numpy): _build_horn_matrix promotes float32 to float64 via untyped np.eye(3)
## Problem

`np.eye(3)` at `src/kabsch_horn/numpy/horn_quat_3d.py:26` returns a float64 array. When `H` is float32, the expression `S - tr[:, np.newaxis, np.newaxis] * I3` on line 30 promotes the entire N matrix to float64. This propagates through `eigh`, `_quat_to_rotation`, and all downstream computations.

Both `horn` and `horn_with_scale` silently return float64 outputs from float32 inputs, while `kabsch` with the same float32 inputs correctly returns float32.

## Suggested Fix

```python
I3 = np.eye(3, dtype=H.dtype)
```

Found by automated audit.

### LINKED ISSUE #128: fix(numpy): no float16 upcast guard -- raw TypeError on float16 inputs
## Problem

All four autodiff frameworks upcast float16/bfloat16 inputs to float32 before calling linear algebra routines, then downcast on output. NumPy'"'"'s `linalg.svd` raises `TypeError: array type float16 is unsupported in linalg` when given float16 input, producing an unhelpful crash.

**Locations:**
- `src/kabsch_horn/numpy/kabsch_svd_nd.py:54` (SVD call)
- `src/kabsch_horn/numpy/horn_quat_3d.py:109` (eigh call)

## Suggested Fix

Add an upcast-on-entry / downcast-on-exit guard matching the pattern in MLX, PyTorch, JAX, and TF. This would make the API consistent across all five frameworks.

Found by automated audit.

### LINKED ISSUE #129: docs(numpy): clarify kabsch vs kabsch_umeyama cross-covariance normalization asymmetry
## Problem

`kabsch` computes `H = p^T @ q` (line 51 of `src/kabsch_horn/numpy/kabsch_svd_nd.py`) while `kabsch_umeyama` computes `H = p^T @ q / N` (line 145). Both are mathematically correct because the SVD directions are scale-invariant; `kabsch_umeyama` needs the normalization for its scale computation.

This is consistent across all frameworks, but the asymmetry could confuse readers without a clarifying comment.

## Suggested Fix

Add a brief inline comment in `kabsch` explaining why normalization is unnecessary, and/or in `kabsch_umeyama` explaining why it normalizes.

Found by automated audit.

### LINKED ISSUE #130: fix: horn/horn_with_scale missing N>=2 validation across all frameworks
## Problem

`kabsch` and `kabsch_umeyama` validate `P.shape[-2] < 2` and raise `ValueError` across all frameworks. The `horn` and `horn_with_scale` functions lack this check in PyTorch, JAX, TensorFlow, and MLX.

With a single point (N=1), centroid subtraction produces all-zero centered points, yielding an all-zero cross-covariance `H` and a degenerate eigendecomposition. `eigh` on this will return arbitrary eigenvectors, producing a meaningless rotation silently.

**Affected locations:**
- `src/kabsch_horn/pytorch/horn_quat_3d.py` -- `horn` and `horn_with_scale`
- `src/kabsch_horn/jax/horn_quat_3d.py` -- `horn` and `horn_with_scale`
- `src/kabsch_horn/tensorflow/horn_quat_3d.py` -- `horn` and `horn_with_scale`
- `src/kabsch_horn/mlx/horn_quat_3d.py` -- `horn` and `horn_with_scale`

(NumPy horn already has this check.)

## Suggested Fix

Add `if P.shape[-2] < 2: raise ValueError("At least 2 points are required for alignment")` to all eight functions, matching the Kabsch implementations.

Found by automated audit.

### LINKED ISSUE #134: TF validation silently skips under tf.function with dynamic shapes
## Problem

All TensorFlow validation checks in `kabsch_svd_nd.py` and `horn_quat_3d.py` use static shape inspection (`P.shape[-2]`, `P.shape[-1]`). Under `tf.function` tracing with dynamic dimensions (e.g., `tf.TensorSpec([None, 3])`), these shape elements are `None`.

Two consequences:

1. **N>=2 checks are silently skipped.** The `is not None` guard means a single-point tensor `[1, 3]` passes validation and produces garbage results without error.

2. **`P.shape[-1] != 3` rejects valid inputs.** `None != 3` evaluates to `True`, so Horn functions incorrectly raise `ValueError` when D is symbolic.

### Affected functions

- `src/kabsch_horn/tensorflow/kabsch_svd_nd.py`: `kabsch()`, `kabsch_umeyama()`
- `src/kabsch_horn/tensorflow/horn_quat_3d.py`: `horn()`, `horn_with_scale()`

### Suggested fix

Add runtime assertions after the static checks:

```python
tf.debugging.assert_greater_equal(
    tf.shape(P)[-2], 2, message="At least 2 points are required for alignment"
)
```

And for the 3D check in Horn:

```python
tf.debugging.assert_equal(
    tf.shape(P)[-1], 3, message="Horn's method is strictly for 3D point clouds"
)
```

These fire during eager execution inside `tf.function` and produce clear errors. The static checks can remain as a fast path for known shapes.

Found during review of #133.

### LINKED ISSUE #137: Consider switching RMSD sqrt guard from max(mse, eps) to mse + eps for smoother gradients
## Summary

The RMSD computation in all autodiff frameworks uses `sqrt(max(mse, eps))` to prevent `d/dx sqrt(0) = inf` in the backward pass. This works, but creates a **gradient discontinuity** at `mse = eps`:

- `mse < eps`: gradient = 0 (clamp active, mse ignored)
- `mse >= eps`: gradient = `1/(2*sqrt(mse))`

At the boundary, the gradient jumps from 0 to `1/(2*sqrt(eps))` (~1,581 for float32). This can cause oscillation during training when the loss is near convergence.

## Proposal

Switch to `sqrt(mse + eps)`, the standard deep learning pattern (used in LayerNorm, GroupNorm, L2 normalization, etc.):

```python
# Current
rmsd = sqrt(max(mse, eps))

# Proposed
rmsd = sqrt(mse + eps)
```

### Benefits
- **Smooth gradient everywhere**: `1/(2*sqrt(mse + eps))` -- no discontinuity
- **Same floor**: minimum RMSD = `sqrt(eps)` (identical to current)
- **Same gradient ceiling**: max gradient = `1/(2*sqrt(eps))` (identical to current)
- **Always nonzero gradient**: even below the floor, gradient responds to mse changes (proportionally tiny, but continuous)

### Tradeoffs
- Slight upward bias on forward RMSD for all values (adds eps inside sqrt). For float32, `sqrt(1.0 + 1.19e-7) - sqrt(1.0) ≈ 6e-8` -- negligible in practice.
- Behavioral change to every RMSD output, so needs test validation

## Affected files (16 sites)

All `kabsch`, `kabsch_umeyama`, `horn`, and `horn_with_scale` functions in:
- `pytorch/kabsch_svd_nd.py`, `pytorch/horn_quat_3d.py`
- `jax/kabsch_svd_nd.py`, `jax/horn_quat_3d.py`
- `tensorflow/kabsch_svd_nd.py`, `tensorflow/horn_quat_3d.py`
- `mlx/kabsch_svd_nd.py`, `mlx/horn_quat_3d.py`

NumPy is unaffected (forward-only, uses `a_min=0.0`).

## Context

Identified during PR #136 review. The `max` clamp was inherited from the original `1e-12` hardcoded value and carried forward into the dtype-aware `finfo(dtype).eps` refactor.

hunter-heidenreich:
### Fixes
- **fix(rmsd)**: switch RMSD sqrt guard from `sqrt(max(mse, eps))` to `sqrt(mse + eps)` across all autodiff frameworks (PyTorch, JAX, TensorFlow, MLX) -- eliminates gradient discontinuity at `mse=eps`, providing smooth gradients everywhere (#138, fixes #137)
- **fix(eps)**: replace all ~46 hardcoded `eps=1e-12` with dtype-aware `finfo(dtype).eps` across all 10 framework source files -- 1e-12 was below float32 machine eps (~1.19e-7) and underflowed to 0 for float16; remove `eps` parameter from `SafeSVD`/`SafeEigh` (PyTorch) and `safe_svd`/`safe_eigh` (JAX) since eps is now internally derived; MLX uses a `_DTYPE_EPS` lookup dict consolidated in `mlx/_utils.py` (#136, fixes #110, fixes #125)
- **fix(tensorflow)**: add `tf.debugging.assert_*` runtime checks alongside static shape validation in all 4 TF public functions so that N>=2, shape-match, and Horn D==3 constraints are enforced under `tf.function` with dynamic shapes; guard Horn's `P.shape[-1] != 3` static check with `is not None` to stop it from incorrectly rejecting valid dynamic-shape inputs (#135, fixes #134)
- **fix(numpy)**: `np.eye(3)` in `_build_horn_matrix` defaulted to float64, silently upcasting float32 inputs; fixed by passing `dtype=H.dtype` (#133, fixes #127)
- **fix(numpy)**: `kabsch`, `kabsch_umeyama`, `horn`, and `horn_with_scale` threw `TypeError` on float16 inputs; added upcast-to-float32/downcast-on-output guard matching other frameworks (#133, fixes #128)
- **fix(horn)**: `horn`/`horn_with_scale` missing N>=2 validation in PyTorch, JAX, TensorFlow, and MLX; added consistent `ValueError` check matching NumPy (#133, fixes #130)
- **fix(mlx)**: removed duplicate `P.shape != Q.shape` checks in `horn`/`horn_with_scale` (dead code) (#133, fixes #120)
- **fix(tensorflow)**: added `is not None` guard to `kabsch`/`kabsch_umeyama` N>=2 checks, preventing `TypeError` under `tf.function` with dynamic shapes (#133)
- **fix(pytorch,tensorflow)**: replace in-place `.diagonal().fill_()`/`.diagonal().zero_()` in PyTorch `SafeSVD`/`SafeEigh` backward with out-of-place `torch.where`, fixing `torch.compile` graph breaks and `create_graph=True` errors; replace `tf.rank(P) == 2` with `len(P.shape) == 2` in TF `horn`/`horn_with_scale` to fix `@tf.function` tracing (#132, fixes #111, fixes #124)
- **fix(autograd)**: TF `safe_svd` backward used `eye * 1e-12` which was zero on off-diagonal entries where `S_i ≈ S_j`, causing `1/0 = inf`; PyTorch `SafeSVD` NaN fallback returned wrong-shaped tensors for rectangular inputs; `nan_s` shape hardened to `min(M, N)` (#131, fixes #123, fixes #116)
- **fix(autograd)**: replace `eps * sign(D + eps)` with `where(D >= 0, eps, -eps)` in all framework backward passes (`kabsch_svd_nd` and `horn_quat_3d`) to eliminate sub-eps sign bias (#131)
- **fix(horn/kabsch)**: add arbitrary batch-dim support to NumPy `horn`/`horn_with_scale` (`[..., N, 3]`), add `N >= 2` and shape-match validation across all frameworks, fix RMSD floor to `0.0`, and clip RMSD before `sqrt` in NumPy Kabsch (#108)
- **fix(jax)**: add shape validation to `horn`/`horn_with_scale`, extract `_horn_core` helper to eliminate ~60 lines of duplication, and clean up idioms (#109)
- **fix(jax)**: replace `vmap_diag` Python conditional in SVD backward with `eye`-broadcast, eliminating retrace on batched/unbatched switch (#107, fixes #31)

### Performance
- **perf**: replace repeated `stack([ones] * (D-1) + [d_sign])` pattern with single `ones()` allocation + index assignment (NumPy, PyTorch), direct `tf.ones(shape)` + `tf.concat` (TensorFlow), and `diag` variable reuse (MLX); JAX already used `.at[..., -1].set()` (#143, fixes #30)

### Refactoring
- **refactor(pytorch)**: remove unreachable `ndim==2` branch in `safe_svd` -- both callers always pass 3D+ tensors (#140, fixes #117)
- **refactor(pytorch)**: add `ctx.needs_input_grad[0]` early return in `SafeSVD.backward` and `SafeEigh.backward` to skip computation when input doesn't require gradients (#140, fixes #115)
- **refactor(jax)**: replace `type(x) is SymbolicZero` with `isinstance(x, SymbolicZero)` in all 5 custom VJP backward checks (#140, fixes #119)
- **refactor(mlx)**: remove unnecessary `mx.array()` scalar construction for division -- MLX supports array/scalar division natively (#140, fixes #122)
- **refactor(mlx)**: remove `mx.array(P)`/`mx.array(Q)` input wrapping from `horn`/`horn_with_scale` for consistency with `kabsch`/`kabsch_umeyama` and other frameworks (#140, fixes #121)

### Documentation
- **docs**: add MkDocs Material documentation site with auto-generated API reference from docstrings (griffe AST parsing, no ML framework imports needed), algorithm pages with LaTeX rendering, and GitHub Actions CI/CD for GitHub Pages (#142)
- **docs**: add/expand docstrings across all frameworks -- Google-style with `[..., N, D]` shape notation, normalization notes for RMSD, and `torch.compile`/`jax.jit` guidance (#141, fixes #27, fixes #28, fixes #126, fixes #129)
- **fix(pytorch)**: correct SVD backward formula comment sign convention (`+ J S + S K` -> `- J S - S K`) to match the code (#139, fixes #112)
- **fix(pytorch)**: expand `SafeSVD` and `SafeEigh` docstrings to document higher-order gradient (`create_graph=True`) support and remove stale "V not V^T" note (#139, fixes #113)
- **fix(pytorch)**: replace deprecated `torch.det` with `torch.linalg.det` in `kabsch` and `kabsch_umeyama` (#139, fixes #114)
- **docs**: document cross-covariance conditioning requirement for rotation stability (#105, fixes #90)

### Features
- **feat(mlx)**: warn when float64 silently falls back to CPU (#103, fixes #51)

### Tests
- **test(tensorflow)**: 20 new tests for dynamic-shape validation under `tf.function` covering N=1 rejection, Horn dim=3 acceptance, shape mismatches, and eager-mode regressions (#135)
- **test**: NumPy N=1 single-point rejection, float16 upcast (single, batched, N-D), float32 dtype preservation for all algorithms (#133)
- **test**: extended `TestSinglePoint` to cover `horn`/`horn_with_scale` across all framework adapters (#133)
- **test(horn)**: add P!=Q near-collinear gradient stability test (#104, fixes #92)

### CI
- **ci**: add docs workflow -- builds on PR to main/dev, deploys to GitHub Pages on push to main (#142)
- **ci**: pin action SHAs, add coverage artifacts, fix fail-fast and setup-python (#101, fixes #5, fixes #6, fixes #7, fixes #8, fixes #9)
- **fix(tests)**: guard `nearly_coplanar_nd` against dim < 3 (#102, fixes #93)

## Test plan

- [x] Full test suite: 6916 passed, 440 skipped, 4 xfailed, 0 failed
- [x] All frameworks (NumPy, PyTorch, JAX, TensorFlow, MLX) green
- [x] `mkdocs build --strict` passes
- [ ] After merge + GitHub Pages enabled, verify docs deploy at https://hunter-heidenreich.github.io/Kabsch-Cookbook/


hunter-heidenreich:
## API Consistency Review

Cross-framework API surface is exceptionally well-aligned in this PR. All 10 source files were audited for signature parity, export consistency, validation logic, return shapes, and dtype handling. No blocking issues found.

### Issues

#### [NITPICK] MLX horn validation order differs from MLX kabsch

In `src/kabsch_horn/mlx/horn_quat_3d.py:65`, `_warn_if_float64(P, Q)` is called **before** the `P.shape[-1] != 3` check, while in `src/kabsch_horn/mlx/kabsch_svd_nd.py:107`, `_warn_if_float64` is called **after** the `dim != 3` and `N < 2` checks. This means a user passing float64 inputs with the wrong dimension to `horn()` will see a float64 warning before the `ValueError`, but the same user calling `kabsch()` will see the `ValueError` without the warning. Not a functional issue -- just a minor ordering inconsistency within MLX.

### Consistency Audit Summary

| Aspect | Status | Notes |
|--------|--------|-------|
| **Function signatures** (`kabsch`, `kabsch_umeyama`, `horn`, `horn_with_scale`) | Consistent | All 5 frameworks: `(P, Q) -> (R, t, rmsd)` or `(R, t, c, rmsd)` |
| **RMSD wrappers** (`kabsch_rmsd`, `kabsch_umeyama_rmsd`) | Consistent | Present in PyTorch, JAX, TF, MLX; correctly absent from NumPy (forward-only) |
| **Framework `__init__.py` exports** | Consistent | NumPy exports 4; autodiff frameworks export 6 each |
| **Top-level re-exports** (`__init__.py`) | Consistent | All suffixed names present (`_numpy`, `_torch`, `_jax`, `_tf`, `_mlx`), silent `ImportError` fallback |
| **Return shapes** (`R`, `t`, `c`, `rmsd`) | Consistent | All frameworks return `[..., D, D]`, `[..., D]`, `[...]`, `[...]` including for unbatched inputs |
| **Input validation** (shape, N>=2, dim==3) | Consistent | All use `ValueError` with identical messages; TF additionally uses `tf.debugging.assert_*` for graph mode |
| **N>=2 check in horn/horn_with_scale** | Consistent | Present in all 5 frameworks (added by this PR for PyTorch, JAX, TF, MLX) |
| **float16/bfloat16 upcasting** | Consistent | All autodiff frameworks upcast both; NumPy upcasts float16 only (no native bfloat16 -- correct) |
| **dtype-aware eps** | Consistent | All use `finfo(dtype).eps` or equivalent (`_DTYPE_EPS` dict for MLX, which lacks `finfo`) |
| **RMSD sqrt guard** | Consistent | NumPy: `sqrt(clip(mse, 0))` (forward-only); autodiff: `sqrt(mse + eps)` (gradient-safe) |
| **Determinant sign at d=0** | Consistent | All produce `+1`; implementation varies (sign+eps, where, direct compare) -- semantically identical |
| **Batch dim handling** | Consistent | NumPy/PyTorch/JAX/TF use `is_single` + reshape; MLX uses broadcasting throughout -- output shapes match |

### Verdict
APPROVED

PR #108, opened 2026-03-09
fix(horn/kabsch): batch dims, cross-framework validation, and RMSD consistency

hunter-heidenreich:
## Summary

**NumPy \`horn\` / \`horn_with_scale\`**
- Add arbitrary batch-dim support (\`[..., N, 3]\`) using the flatten/reshape pattern already used by \`kabsch\`; previously only \`[N, 3]\` and \`[B, N, 3]\` worked correctly
- Add \`P.shape == Q.shape\` and \`N >= 2\` validation; replace bare \`assert\` with \`raise ValueError\` for the 3D check so it survives \`python -O\`
- Add Google-style docstrings
- Extract \`_build_horn_matrix\` and \`_quat_to_rotation\` private helpers, eliminating copy-pasted 4x4 matrix construction and quaternion-to-R expansion
- Replace \`np.trace(H, axis1=1, axis2=2)\` with \`H.diagonal(axis1=-2, axis2=-1).sum(-1)\`

**NumPy \`kabsch\` / \`kabsch_umeyama\`**
- Add \`np.clip(..., a_min=0.0)\` before \`np.sqrt\` in RMSD computation to prevent NaN from tiny negative rounding errors
- Add \`N >= 2\` validation
- Fix dimension comments (\`Bx1x3\` -> \`Bx1xD\` etc.); replace in-place \`d_sign\` mutation with \`np.where\`

**All autodiff frameworks (PyTorch, JAX, TensorFlow, MLX) -- \`horn\` / \`horn_with_scale\`**
- Add \`P.shape == Q.shape\` validation (was missing entirely)
- Replace bare \`assert\` with \`raise ValueError\` for the 3D check in PyTorch (was silently stripped under \`python -O\`); add explicit 3D check to JAX, TF, and MLX
- Set RMSD floor to \`0.0\` (was \`1e-12\`, causing \`horn(P, P)\` to return RMSD ~= 1e-6 instead of 0)

**All frameworks -- \`kabsch\` / \`kabsch_umeyama\`**
- Add \`N >= 2\` validation across NumPy, PyTorch, JAX, TensorFlow, and MLX

## Test plan

- [x] \`uv run pytest tests/ -x -q\` -- 6814 passed, 440 skipped, 4 xfailed
- [x] \`uv run ruff check src/kabsch_horn/\` -- no errors
- [x] Manual smoke test: \`horn(P, P)\` with \`P.shape == (2, 3, 10, 3)\` returns \`R.shape == (2, 3, 3, 3)\` and near-zero RMSD (\`< 1e-10\`); \`horn_with_scale\` likewise; shape mismatch, non-3D, and N=1 all raise \`ValueError\`

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -167,7 +167,7 @@ def horn(
     rmsd = jnp.sqrt(
         jnp.clip(
             jnp.sum(jnp.square(aligned - q), axis=(1, 2)) / N_pts,
-            min=0.0,
+            min=1e-12,
             max=None,
         )
     )
@@ -248,7 +248,7 @@ def horn_with_scale(
     )
     diff = aligned_P - Q
     rmsd = jnp.sqrt(
-        jnp.clip(jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts, min=0.0, max=None)
+        jnp.clip(jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts, min=1e-12, max=None)
     )
 
     if is_single:
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -139,7 +139,8 @@ def horn(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
-
+    if P.shape[-1] != 3:
+        raise ValueError("Horn's method is strictly for 3D point clouds")
     orig_dtype = P.dtype
     if orig_dtype in (jnp.float16, jnp.bfloat16):
         P = P.astype(jnp.float32)
@@ -211,7 +212,8 @@ def horn_with_scale(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
-
+    if P.shape[-1] != 3:
+        raise ValueError("Horn's method is strictly for 3D point clouds")
     orig_dtype = P.dtype
     if orig_dtype in (jnp.float16, jnp.bfloat16):
         P = P.astype(jnp.float32)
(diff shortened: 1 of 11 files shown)


hunter-heidenreich:
## NumPy Review

Clean, well-reasoned PR. The batch-dim reshape pattern is correctly applied to `horn`/`horn_with_scale`, the helper extraction eliminates real duplication, and the numerical fixes (`np.where` for `d_sign`, `np.clip` before `sqrt`) are all correct.

### Issues

#### [NITPICK] `_build_horn_matrix` docstring documents `(B, 3, 3)` but uses positional `(0, 2, 1)` transpose internally

`horn_quat_3d.py` lines 13 and 30 use `H.transpose(0, 2, 1)` and `Delta[:, :, np.newaxis]` -- both assume axis 0 is the batch dimension. This is always true as called (H is always flattened to `(B, 3, 3)` before being passed in), but a caller reading the helper in isolation would not know it. The docstring already says `(B, 3, 3)`, so this is just a note: if the helper ever gets reused with a differently-shaped H the positional transposes will silently produce wrong output instead of an indexing error.

#### [NITPICK] `is_single` path computes `batch_dims` that it never uses

In `horn` and `horn_with_scale`, after the `is_single` newaxis the shape becomes `(1, N, 3)`, so `batch_dims = (1,)` is computed on line 77 (and 137). That value is never reached because the `if is_single` early return fires first. This is harmless but slightly confusing -- the `batch_dims` variable appears to be needed, but the single-sample return path bypasses it entirely. No action required; noting for clarity.

#### [NITPICK] No `N >= 2` guard on `horn` or `horn_with_scale`

With `N=1`, centering produces an all-zero `p`/`q`, making `H` a zero matrix. `eigh` on a zero 4x4 matrix returns an arbitrary eigenvector, meaning `R` is numerically undefined (though RMSD will be 0). The same pre-existing gap exists in `kabsch`. Since this PR adds validation elsewhere, it would be a good opportunity to add a matching `N >= 2` check to `horn`/`horn_with_scale` as well (this is not a regression, just a gap).

### Verdict

APPROVED

PR #109, opened 2026-03-11
fix(jax): add shape validation to horn, extract _horn_core, and clean up idioms

hunter-heidenreich:
## Summary

- **Bug fix**: add `P.shape != Q.shape` guard to `horn` and `horn_with_scale` -- this was missing unlike `kabsch`/`kabsch_umeyama` which already had the check; makes the full API consistent
- **Deduplication**: extract `_horn_core` helper to eliminate ~60 lines of verbatim code shared between `horn` and `horn_with_scale`
- **Idiom fixes**: replace `vmap_diag` conditional with broadcasting in `_eigh_bwd`; replace `[ones]*(D-1)+[d_sign]` list construction with `.at[..., -1].set()` in `kabsch` and `kabsch_umeyama`
- **Efficiency**: cache `diag_mask` in `_eigh_bwd` and `_bwd` instead of calling `jnp.eye` twice per backward pass
- **Clarity**: add comment on the `sign(d + 1e-12)` epsilon nudge; note `safe_svd` `V` vs `Vh` return convention in docstring
- **Tests**: extend `test_raises_error_when_dims_differ` to cover `horn` and `horn_with_scale` (was `kabsch`/`umeyama` only)

## Test plan

- [x] `uv run pytest tests/test_error_handling.py` -- 184 passed, 8 skipped (new horn shape-mismatch cases pass)
- [x] `uv run pytest tests/test_differentiability_traps.py tests/test_gradient_verification.py tests/test_degeneracy.py` -- 5020 passed, 384 skipped, 4 xfailed
- [x] `uv run ruff check src/kabsch_horn/jax/ && uv run ruff format --check src/kabsch_horn/jax/` -- clean


hunter-heidenreich:
## pytest Review

The only pytest-facing change is extending a `@pytest.mark.parametrize` list by two string values. The idiom is correct and consistent with the rest of the class.

### Issues

#### [NITPICK] `pytest.raises` missing `match=` pattern

`tests/test_error_handling.py` lines 29 and 51 -- both `pytest.raises` calls lack a `match=` argument. The pytest docs treat `match=` as a best-practice because it confirms the right code path raised the exception, not just any code that happens to throw the same type. A minimal pattern like `match=r"same shape"` would be sufficient.

This predates the PR (the existing `kabsch`/`umeyama` cases have the same omission), so it is not introduced here. The new `horn`/`horn_with_scale` entries inherit the same gap.

### Verdict

APPROVED

PR #131, opened 2026-03-12
fix(autograd): gradient correctness, NaN shape hardening, and sign-symmetric safe denominators

### LINKED ISSUE #116: fix(pytorch): SafeSVD NaN-fallback saves wrong-shape tensor for Vh

### LINKED ISSUE #123: fix(tensorflow): safe_svd backward eye-based masking produces inf/NaN for near-equal singular values

hunter-heidenreich:
## Summary

- **TF `safe_svd` backward** (#123): The `eye * 1e-12` mask was zero on off-diagonal entries where `S_i ≈ S_j`, causing `1/0 = inf`. Replaced with sign-preserving `where(D >= 0, eps, -eps)` pattern.
- **PyTorch `SafeSVD` NaN fallback** (#116): `nan_mat` (shape `A.shape`) was reused for both U and Vh in `save_for_backward`. For rectangular inputs Vh is `[..., N, N]`, not `A.shape`. Now creates properly-shaped `nan_u` and `nan_vh` tensors, and returns `nan_vh` (not `nan_u`) for V.
- **`nan_s` shape hardening**: `nan_s` now uses `min(M, N)` instead of `A.shape[:-1]`, matching the real SVD contract for non-square inputs.
- **Sign-symmetric safe denominators**: Replaced `eps * sign(D + eps)` with `where(D >= 0, eps, -eps)` across all framework backward passes (`kabsch_svd_nd` and `horn_quat_3d` in PyTorch, JAX, TensorFlow, and MLX). The old pattern biased sub-eps negative differences to `+eps` because the `+eps` nudge inside `sign()` made them positive.

Fixes #123, fixes #116.

## Test plan

- [x] `uv run pytest tests/ -v` -- full suite: 6846 passed, 440 skipped, 0 failures

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -37,7 +37,7 @@ def _eigh_bwd(res: tuple, g: tuple) -> tuple:
     D = L[..., jnp.newaxis, :] - L[..., jnp.newaxis]
 
     mask = jnp.abs(D) < eps
-    safe_D = jnp.where(mask, eps * jnp.sign(D + eps), D)
+    safe_D = jnp.where(mask, jnp.where(D >= 0, eps, -eps), D)
     diag_mask = jnp.eye(D.shape[-1], dtype=bool)
     safe_D = jnp.where(diag_mask, 1.0, safe_D)
     F = jnp.where(diag_mask, 0.0, 1.0 / safe_D)
--- a/src/kabsch_horn/jax/kabsch_svd_nd.py
+++ b/src/kabsch_horn/jax/kabsch_svd_nd.py
@@ -61,7 +61,7 @@ def _bwd(res, g):
     D = S_sq[..., jnp.newaxis] - S_sq[..., jnp.newaxis, :]  # BxDxD
 
     # 3. Safe F
-    safe_D = jnp.where(jnp.abs(D) < eps, eps * jnp.sign(D + eps), D)
+    safe_D = jnp.where(jnp.abs(D) < eps, jnp.where(D >= 0, eps, -eps), D)
     diag_mask = jnp.eye(D.shape[-1], dtype=bool)
     safe_D = jnp.where(diag_mask, 1.0, safe_D)
     F = jnp.where(diag_mask, 0.0, 1.0 / safe_D)
--- a/src/kabsch_horn/mlx/horn_quat_3d.py
+++ b/src/kabsch_horn/mlx/horn_quat_3d.py
@@ -23,7 +23,7 @@ def safe_eigh_bwd(primals, cotangents, outputs):
     eye = mx.eye(D.shape[-1], dtype=D.dtype)
     eps = 1e-12
     mask = mx.abs(D) < eps
-    safe_D = mx.where(mask, eps * mx.sign(D + eps), D)
+    safe_D = mx.where(mask, mx.where(D >= 0, eps, -eps), D)
     # Set diagonal to 1.0 so 1/safe_D is defined everywhere; zero it out after
     safe_D = mx.where(eye == 1, mx.ones_like(safe_D), safe_D)
 
--- a/src/kabsch_horn/pytorch/horn_quat_3d.py
+++ b/src/kabsch_horn/pytorch/horn_quat_3d.py
@@ -27,7 +27,7 @@ class SafeEigh(torch.autograd.Function):
 
         # 2. Mask unstable divides directly
         mask = torch.abs(D) < eps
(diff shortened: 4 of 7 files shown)


hunter-heidenreich:
## PyTorch Review

The PR fixes `SafeSVD.forward` NaN-fallback shapes to match the real SVD output contract. The shapes are now correctly documented and constructed. However, there is one bug on the return line and one latent shape issue worth noting.

### Issues

#### [BLOCKING] Return uses `nan_u` instead of `nan_vh` as third value (line 29)

`src/kabsch_horn/pytorch/kabsch_svd_nd.py:29`

```python
return nan_u, nan_s, nan_u   # <-- third value should be nan_vh (or nan_v)
```

The normal code path returns `(U, S, V)` where `V = Vh.mH` (line 32-36). The NaN fallback should mirror this and return a NaN tensor with V's shape (`[..., N, N]` for full_matrices), not U's shape (`[..., M, M]`).

In practice this is **currently harmless** because `safe_svd` is only ever called on the square cross-covariance matrix H `[B, D, D]`, so M == N and `nan_u` and `nan_vh` have identical shapes. The values are all NaN regardless. But:
- It contradicts the comment on line 23 (`# U is [..., M, M], ... Vh is [..., N, N]`) that was added in this very PR.
- If `SafeSVD` is ever called on a non-square matrix, this would silently return a wrong-shaped tensor.
- `save_for_backward` correctly saves `nan_vh` (line 27), making the asymmetry between saved and returned tensors confusing.

Fix: change line 29 to:
```python
nan_v = nan_vh  # all-NaN, transpose is identity on content
return nan_u, nan_s, nan_v
```
Or simply `return nan_u, nan_s, nan_vh` since for all-NaN tensors the `.mH` distinction is meaningless.

#### [RECOMMENDED] `nan_s` shape is `[..., M]` not `[..., min(M, N)]` (line 25)

`src/kabsch_horn/pytorch/kabsch_svd_nd.py:25`

```python
nan_s = A.new_full(A.shape[:-1], float("nan"))
```

`A.shape[:-1]` gives `[..., M]`. The real SVD returns S with shape `[..., min(M, N)]`. For the square-matrix-only usage in this codebase, M == N so this is correct. But since the comment on line 23 documents the general contract, the shape should match:

```python
nan_s = A.new_full((*A.shape[:-2], min(A.shape[-2], A.shape[-1])), float("nan"))
```

This is a low-priority hardening item given the square-only call sites.

#### [NITPICK] In-place `.fill_` / `.zero_` in backward (lines 75, 79)

Lines 75 and 79 use in-place ops on `safe_D.diagonal(...)`. This works because `safe_D` is a fresh tensor from `torch.where`, so it's safe for autograd. No action needed, just noting this was checked.

### Verdict
CHANGES REQUESTED

The `nan_u`-vs-`nan_vh` return mismatch on line 29 is a clear copy-paste bug. While it is currently benign due to square-only call sites, it directly contradicts the shape comments added in this same PR and would break for any future non-square usage. Please fix before merging.

PR #132, opened 2026-03-12
fix(pytorch,tensorflow): remove in-place mutations and tf.rank tracing bug

### LINKED ISSUE #111: fix(pytorch): in-place view mutations in SafeSVD/SafeEigh backward block higher-order gradients and torch.compile

### LINKED ISSUE #124: fix(tensorflow): tf.rank() in Python if-statement breaks @tf.function tracing

hunter-heidenreich:
## Summary
- **#111**: Replace in-place `.diagonal().fill_()` / `.diagonal().zero_()` in PyTorch `SafeSVD.backward` and `SafeEigh.backward` with out-of-place `torch.where` using a boolean diagonal mask. Fixes `torch.compile` graph breaks and `RuntimeError` under `create_graph=True`.
- **#124**: Replace `tf.rank(P) == 2` with `len(P.shape) == 2` in TF `horn` and `horn_with_scale`. `tf.rank()` returns a symbolic tensor, so the Python `if` fails under `@tf.function` tracing.

Closes #111
Closes #124

## Test plan
- [x] Full test suite passes (6846 passed, 440 skipped, 4 xfailed)

### DIFF
--- a/src/kabsch_horn/pytorch/horn_quat_3d.py
+++ b/src/kabsch_horn/pytorch/horn_quat_3d.py
@@ -29,10 +29,11 @@ class SafeEigh(torch.autograd.Function):
         mask = torch.abs(D) < eps
         safe_D = torch.where(mask, torch.where(D >= 0, eps, -eps), D)
 
-        # 3. Prevent diagonal inversion problems outright
-        safe_D.diagonal(dim1=-2, dim2=-1).fill_(1.0)
+        # 3. Prevent diagonal inversion problems (out-of-place for torch.compile)
+        diag_mask = torch.eye(safe_D.shape[-1], dtype=torch.bool, device=safe_D.device)
+        safe_D = torch.where(diag_mask, torch.ones_like(safe_D), safe_D)
         F = 1.0 / safe_D
-        F.diagonal(dim1=-2, dim2=-1).zero_()
+        F = torch.where(diag_mask, torch.zeros_like(F), F)
 
         # 4. Standard backprop algebra using safe denominators
         Vt_dV = torch.matmul(V.mH, grad_V)
--- a/src/kabsch_horn/pytorch/kabsch_svd_nd.py
+++ b/src/kabsch_horn/pytorch/kabsch_svd_nd.py
@@ -72,12 +72,13 @@ class SafeSVD(torch.autograd.Function):
         # Safe denominator preserving sign for sub-eps values
         safe_D = torch.where(mask, torch.where(D >= 0, eps, -eps), D)
 
-        # Protect diagonal from 1/0
-        safe_D.diagonal(dim1=-2, dim2=-1).fill_(1.0)
+        # Protect diagonal from 1/0 (out-of-place for torch.compile)
+        diag_mask = torch.eye(safe_D.shape[-1], dtype=torch.bool, device=safe_D.device)
+        safe_D = torch.where(diag_mask, torch.ones_like(safe_D), safe_D)
 
         F = 1.0 / safe_D
         # Set diagonal to exactly 0 to satisfy Hadamard condition
-        F.diagonal(dim1=-2, dim2=-1).zero_()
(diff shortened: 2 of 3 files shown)


hunter-heidenreich:
## Correctness Review

Both changes are pure mechanical refactors with no algorithmic or numerical impact. The backward pass F-matrix construction produces bit-identical results, and the TF rank check is a control-flow fix with no effect on tensor computation.

### Issues

No issues found.

**SafeSVD backward (kabsch_svd_nd.py:75-81)**: The diagonal of `D = S_i^2 - S_j^2` is always zero (since i==j), which gets masked to `eps` by the preceding `torch.where`. The old code then overwrote these diagonal entries with `fill_(1.0)`, computed `F = 1/safe_D` (giving F_ii = 1.0), then zeroed them with `zero_()`. The new code does the same via `torch.where(diag_mask, ones, safe_D)` → `F = 1/safe_D` → `torch.where(diag_mask, zeros, F)`. Off-diagonal elements are never touched by `diag_mask`. The result is numerically identical.

**SafeEigh backward (horn_quat_3d.py:32-36)**: Same pattern as SafeSVD but for eigenvalue differences `D = L_j - L_i`. The diagonal is zero → masked to eps → overwritten to 1.0 → inverted → zeroed. The new code follows the same sequence out-of-place. Numerically identical.

**Cross-framework consistency**: The new PyTorch pattern (`torch.eye` + `torch.where`) aligns with the existing JAX implementation (`jax/kabsch_svd_nd.py:65-67`, `jax/horn_quat_3d.py:41-43`) and the existing TF implementation (`tf.linalg.set_diag`). All three frameworks now use out-of-place diagonal masking.

**TF rank check (horn_quat_3d.py:55,159)**: `len(P.shape) == 2` evaluates the static rank, producing a Python `bool`. This is functionally identical to the old `tf.rank(P) == 2` in eager mode, and correct under `@tf.function` tracing where the old code would fail. No numerical computation is affected.

### Verdict
APPROVED

PR #136, opened 2026-03-13
fix: replace hardcoded eps=1e-12 with dtype-aware finfo(dtype).eps (#110)

### LINKED ISSUE #110: refactor: replace hardcoded eps=1e-12 with dtype-aware floor

hunter-heidenreich:
## Summary
- Replaces all ~46 hardcoded `1e-12` eps values across all 10 framework source files with dtype-derived machine epsilon
- Backward-pass masking (SafeSVD/SafeEigh): eps derived from singular value / eigenvalue tensor dtype via `finfo`
- Forward-pass clamps (RMSD sqrt, scale division, det sign nudge): eps derived from working tensor dtype
- Removes `eps` parameter from PyTorch `SafeSVD`/`SafeEigh` and JAX `safe_svd`/`safe_eigh` since eps is now internally derived
- MLX uses a `_DTYPE_EPS` lookup dict (MLX lacks `finfo`)
- TF/NumPy use `np.finfo` for dtype resolution

| dtype | finfo.eps | old 1e-12 | effect |
|-------|-----------|-----------|--------|
| float16 | 9.77e-4 | underflows to 0 | now functional |
| bfloat16 | 7.81e-3 | underflows to 0 | now functional |
| float32 | 1.19e-7 | too tight | masks float32 noise properly |
| float64 | 2.22e-16 | too loose | more precise masking |

## Test plan
- [x] Full test suite: 6896 passed, 440 skipped, 4 xfailed (after adding mx.float64 to _DTYPE_EPS)
- [x] `ruff check . && ruff format .` clean
- [x] Gradient verification tests pass across all frameworks and dtypes
- [x] MLX-Float64 tests pass (previously failed with missing float64 eps entry)

Closes #110

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -3,21 +3,21 @@ import jax.numpy as jnp
 
 
 @jax.custom_vjp
-def safe_eigh(A: jnp.ndarray, eps: float = 1e-12) -> tuple[jnp.ndarray, jnp.ndarray]:
+def safe_eigh(A: jnp.ndarray) -> tuple[jnp.ndarray, jnp.ndarray]:
     """Gradient-safe eigendecomposition for symmetric matrices. Masks near-zero
     eigenvalue differences (< eps) in the backward pass to prevent NaN gradients."""
     return jnp.linalg.eigh(A)
 
 
 def _eigh_fwd(
-    A: jnp.ndarray, eps: float
+    A: jnp.ndarray,
 ) -> tuple[tuple[jnp.ndarray, jnp.ndarray], tuple]:
     L, V = jnp.linalg.eigh(A)
-    return (L, V), (L, V, eps)
+    return (L, V), (L, V)
 
 
 def _eigh_bwd(res: tuple, g: tuple) -> tuple:
-    L, V, eps = res
+    L, V = res
     grad_L, grad_V = g
 
     grad_L = (
@@ -36,6 +36,7 @@ def _eigh_bwd(res: tuple, g: tuple) -> tuple:
 
     D = L[..., jnp.newaxis, :] - L[..., jnp.newaxis]
 
+    eps = jnp.finfo(L.dtype).eps
     mask = jnp.abs(D) < eps
     safe_D = jnp.where(mask, jnp.where(D >= 0, eps, -eps), D)
     diag_mask = jnp.eye(D.shape[-1], dtype=bool)
@@ -49,7 +50,7 @@ def _eigh_bwd(res: tuple, g: tuple) -> tuple:
     term = L_diag + F * (Vt_dV - mH(Vt_dV)) / 2
     grad_A = jnp.matmul(V, jnp.matmul(term, mH(V)))
 
-    return (grad_A, None)
+    return (grad_A,)
 
 
 safe_eigh.defvjp(_eigh_fwd, _eigh_bwd)
@@ -166,10 +167,11 @@ def horn(
     )
 
     aligned = jnp.matmul(p, jnp.swapaxes(R, 1, 2))
+    _eps = jnp.finfo(P.dtype).eps
     rmsd = jnp.sqrt(
         jnp.clip(
             jnp.sum(jnp.square(aligned - q), axis=(1, 2)) / N_pts,
-            min=1e-12,
+            min=_eps,
             max=None,
         )
     )
@@ -238,8 +240,9 @@ def horn_with_scale(
 
     # H is unscaled (p.T @ q); var_P * N_pts = sum(sq(p)), so c = tr(R^T H) / sum(sq(p))
     var_P = jnp.sum(jnp.square(p), axis=(1, 2)) / N_pts
+    _eps = jnp.finfo(P.dtype).eps
     c = jnp.sum(R * jnp.swapaxes(H, 1, 2), axis=(1, 2)) / (
-        jnp.clip(var_P, min=1e-12) * N_pts
+        jnp.clip(var_P, min=_eps) * N_pts
     )
 
     t = jnp.squeeze(centroid_Q, axis=1) - c[:, jnp.newaxis] * jnp.squeeze(
@@ -252,7 +255,7 @@ def horn_with_scale(
     )
     diff = aligned_P - Q
     rmsd = jnp.sqrt(
-        jnp.clip(jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts, min=1e-12, max=None)
+        jnp.clip(jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts, min=_eps, max=None)
     )
 
     if is_single:
--- a/src/kabsch_horn/jax/kabsch_svd_nd.py
+++ b/src/kabsch_horn/jax/kabsch_svd_nd.py
@@ -4,7 +4,7 @@ import jax.numpy as jnp
 
 @jax.custom_vjp
 def safe_svd(
-    A: jnp.ndarray, eps: float = 1e-12
+    A: jnp.ndarray,
 ) -> tuple[jnp.ndarray, jnp.ndarray, jnp.ndarray]:
     """Gradient-safe SVD. Masks near-zero singular value differences (< eps) in the
     backward pass to prevent NaN gradients for symmetric or degenerate inputs.
@@ -19,17 +19,17 @@ def safe_svd(
 
 
 def _fwd(
-    A: jnp.ndarray, eps: float
+    A: jnp.ndarray,
 ) -> tuple[tuple[jnp.ndarray, jnp.ndarray, jnp.ndarray], tuple]:
     U, S, Vh = jnp.linalg.svd(A)
     # return primal outputs and residuals for backward pass
     V = jnp.swapaxes(jnp.conj(Vh), -1, -2)
-    return (U, S, V), (U, S, Vh, eps)
+    return (U, S, V), (U, S, Vh)
 
 
 def _bwd(res, g):
     # Retrieve
-    U, S, Vh, eps = res
+    U, S, Vh = res
     grad_U, grad_S, grad_V = g
 
     # Check if None or SymbolicZero
@@ -61,6 +61,7 @@ def _bwd(res, g):
     D = S_sq[..., jnp.newaxis] - S_sq[..., jnp.newaxis, :]  # BxDxD
 
     # 3. Safe F
+    eps = jnp.finfo(S.dtype).eps
     safe_D = jnp.where(jnp.abs(D) < eps, jnp.where(D >= 0, eps, -eps), D)
     diag_mask = jnp.eye(D.shape[-1], dtype=bool)
     safe_D = jnp.where(diag_mask, 1.0, safe_D)
@@ -82,7 +83,7 @@ def _bwd(res, g):
 
     grad_A = jnp.matmul(U, jnp.matmul(term, Vh))
 
-    return grad_A, None
+    return (grad_A,)
 
 
 safe_svd.defvjp(_fwd, _bwd)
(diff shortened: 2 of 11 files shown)


hunter-heidenreich:
## JAX Review

The JAX changes correctly replace hardcoded `1e-12` with dtype-derived `jnp.finfo(...).eps` throughout both `kabsch_svd_nd.py` and `horn_quat_3d.py`. The implementation is clean and consistent across SVD and eigh backward passes as well as all forward-pass clamp/sign sites.

### Issues

#### [NITPICK] `type(x) is SymbolicZero` vs `isinstance`

Lines 37-50 in `kabsch_svd_nd.py` and 23-32 in `horn_quat_3d.py` use `type(grad_X) is jax.custom_derivatives.SymbolicZero` rather than `isinstance`. This is intentional (and correct) since `SymbolicZero` is a sentinel that should not be subclassed, so identity-check is fine. No action needed -- just noting for posterity.

### Detailed Assessment

**1. custom_vjp correctness** -- `safe_svd` now takes a single primal argument `(A,)`. `_fwd` returns `(primals, residuals)` with residuals `(U, S, Vh)`. `_bwd` returns `(grad_A,)` -- a 1-tuple matching the single primal input. Previously with `eps` as a second parameter it returned `(grad_A, None)`. The removal of both the parameter and the `None` sentinel is correctly synchronized. Same story for `safe_eigh` / `_eigh_fwd` / `_eigh_bwd`. All correct.

**2. `jnp.finfo` in `_bwd`** -- `jnp.finfo(S.dtype).eps` and `jnp.finfo(L.dtype).eps` each return a Python float, which becomes a JAX constant when used in arithmetic. Since `_bwd` is called by the custom_vjp machinery (not independently traced by `jit`), there is no abstract-tracing concern. The dtype of `S` / `L` is concrete at backward-pass time because it was materialized during the forward pass and stored in residuals. No issue here.

**3. Functional purity** -- Both backward functions are pure: no in-place mutation, no side effects, no captured mutable state. The forward-pass functions (`kabsch`, `kabsch_umeyama`, `horn`, `horn_with_scale`) use only `jnp` ops and `.reshape` (which returns a view, not a mutation). Clean.

**4. SymbolicZero handling** -- All six gradient slots (grad_U, grad_S, grad_V in SVD; grad_L, grad_V in eigh) are checked for `SymbolicZero` and replaced with `jnp.zeros_like(...)` of the corresponding residual. This is the standard pattern for JAX custom_vjp when some outputs may not be differentiated through. Correct.

**5. Forward-pass eps usage** -- `_eps = jnp.finfo(P.dtype).eps` is used consistently for `jnp.sign(d + _eps)` nudging and `jnp.clip(..., min=_eps)` in RMSD computation across all four public functions. The variable is derived after any float16/bfloat16 upcast, so it always reflects the working precision (float32 or float64). Correct.

### Verdict

APPROVED -- The changes are mechanically sound. The custom_vjp signatures match, `jnp.finfo` usage is safe in the backward context, there are no tracing hazards, and functional purity is maintained throughout.

hunter-heidenreich:
## NumPy Review

The two changes in `numpy/kabsch_svd_nd.py` (`kabsch_umeyama`, line 178) and `numpy/horn_quat_3d.py` (`horn_with_scale`, line 206) correctly replace the hardcoded `1e-12` scale-clamping floor with `np.finfo(P.dtype).eps`. This makes the clamping threshold adapt to the working precision.

### Issues

#### [NITPICK] Consistency note -- no changes needed in `kabsch()` or `horn()`

Confirmed: `kabsch()` and `horn()` do not use any epsilon-based clamping. Their only `np.clip` call is `a_min=0.0` on the RMSD argument to `np.sqrt`, which guards against negative values from floating-point noise. This is semantically different from the scale-denominator clamping and does not need `finfo(dtype).eps`. The PR correctly limits its scope to the two scale-estimation paths.

#### [NITPICK] `np.finfo` and dtype coverage

`np.finfo(P.dtype).eps` is the right API for float32/float64. NumPy does not natively support `bfloat16`, and the existing `float16` upcast-to-`float32` guard ensures `P.dtype` is always a type `np.finfo` can handle by the time the eps lookup runs. No issue here.

### Verdict
APPROVED

hunter-heidenreich:
## TensorFlow Review

The TF changes replace all hardcoded `1e-12` epsilon values with `np.finfo(dtype).eps`, matching the pattern already used in the PyTorch backend (`torch.finfo(S.dtype).eps`). The implementation is correct and well-structured.

### Issues

#### [NITPICK] `safe_eigh` is now a trivial wrapper

After removing the `eps` parameter, `safe_eigh` (line 5-6 of `horn_quat_3d.py`) is just `return tf.linalg.eigh(A)`. It exists solely as an indirection for `call_safe_eigh` to invoke. This is fine for readability/symmetry with other frameworks, but worth noting it carries no logic of its own anymore.

#### [NITPICK] Bonus bug fix in `safe_D` sign logic deserves a mention

The eigh backward changed from:
```python
safe_D = tf.where(mask, eps * tf.sign(D + eps), D)
```
to:
```python
safe_D = tf.where(mask, tf.where(D >= 0, eps, -eps), D)
```

The old code had a subtle sign bug: for small negative `D` values (e.g., `-1e-13`), adding `eps` (`1e-12`) made the sum positive, so `tf.sign` returned `+1` instead of `-1`. The new code correctly preserves the sign for all sub-eps values. This is a genuine correctness fix but it is bundled into a PR titled as a dtype-aware eps change -- if it matters for the changelog, consider calling it out explicitly.

### Analysis of Key Concerns

**`np.finfo` inside `@tf.custom_gradient`**: In `safe_svd` (line 36) and `call_safe_eigh` (line 12), `np.finfo(S.dtype.as_numpy_dtype).eps` is evaluated once when the closure is created during the forward pass. Since `S.dtype` is fixed per trace, this produces a correct constant. Under `@tf.function`, `dtype` is known statically at trace time, so `as_numpy_dtype` resolves without issues. Correct.

**`np.finfo` in forward-pass functions**: The `_eps = np.finfo(P.dtype.as_numpy_dtype).eps` calls produce a Python float that TF auto-promotes to a constant of the matching dtype. Correct and consistent with how PyTorch uses `torch.finfo(...).eps` as a Python-level scalar.

**`import numpy`**: NumPy is a mandatory transitive dependency of TensorFlow. No new dependency introduced.

**Consistency with PyTorch backend**: The pattern is now aligned -- both frameworks derive eps from the tensor's actual dtype rather than assuming float64-scale precision for all types. This matters for float32 inputs where `finfo(float32).eps ~= 1.19e-7` vs the old `1e-12`, giving a more appropriate threshold.

### Verdict

APPROVED

hunter-heidenreich:
## API Consistency Review

This PR removes the `eps` parameter from the `safe_svd` and `safe_eigh` signatures in PyTorch and JAX, deriving eps from `finfo(dtype).eps` internally instead. I reviewed all 10 source files across all 5 frameworks for signature consistency, public API stability, and cross-framework parity.

### Issues

#### [NITPICK] Pre-existing: NumPy `kabsch` uses different eps strategy for RMSD and det sign

NumPy's `kabsch` (line 83-89 of `numpy/kabsch_svd_nd.py`) clips RMSD with `a_min=0.0` rather than `a_min=finfo(eps)`, and uses `np.where(d == 0, 1.0, np.sign(d))` for the determinant sign instead of the `sign(d + eps)` nudge used by PyTorch and JAX. Meanwhile NumPy's `kabsch_umeyama` does use `np.finfo(P.dtype).eps` for the scale denominator clamp (line 178).

This is a pre-existing asymmetry, not introduced by this PR. NumPy is forward-only so the gradient-safety motivation for eps-clamped RMSD sqrt does not apply, and `np.where(d == 0, ...)` is arguably cleaner than the `sign(d + eps)` nudge for forward-only code. No action required from this PR.

#### [NITPICK] Pre-existing: TF and MLX det sign nudge differs from PyTorch/JAX

PyTorch and JAX both use `sign(d + finfo(dtype).eps)` to handle the zero-determinant edge case (lines 189, 295 in PyTorch; lines 153, 251 in JAX). TensorFlow uses `tf.where(d_sign == 0, tf.ones_like(d_sign), d_sign)` (explicit post-hoc fix-up), and MLX uses `mx.where(d < 0, -1.0, 1.0)` (binary branch, no zero case). All three approaches produce the same result for the d=0 edge case (mapping to +1), so this is a stylistic difference only. Pre-existing, not introduced by this PR.

#### [NITPICK] MLX uses a hardcoded `_DTYPE_EPS` dict instead of a runtime `finfo` call

MLX's `kabsch_svd_nd.py` and `horn_quat_3d.py` both use a module-level `_DTYPE_EPS` dictionary (lines 5-9) rather than a runtime `finfo` equivalent. This is a reasonable workaround since MLX does not expose `mx.finfo`, and the values match the IEEE 754 constants. The fallback default (`1.1920929e-7`, i.e. float32 eps) is sensible. Pre-existing, not introduced by this PR.

### Findings (all clean)

**1. safe_svd / safe_eigh signatures -- consistent across frameworks.**

| Framework | `safe_svd` signature | `safe_eigh` signature |
|-----------|---------------------|-----------------------|
| PyTorch | `SafeSVD.forward(ctx, A)` / `safe_svd(A)` | `SafeEigh.forward(ctx, A)` |
| JAX | `safe_svd(A)` / `_fwd(A)` | `safe_eigh(A)` / `_eigh_fwd(A)` |
| TensorFlow | `safe_svd(A)` (via `@tf.custom_gradient`) | `call_safe_eigh(A)` (via `@tf.custom_gradient`) |
| MLX | `safe_svd(A)` (via `@mx.custom_function`) | `safe_eigh_fwd(A)` (via `@mx.custom_function`) |

No framework takes an `eps` parameter in any of these signatures. TF and MLX never had one (confirmed); PyTorch and JAX had it removed by this PR. All four frameworks now derive eps internally from dtype metadata.

**2. Public API unchanged.** `kabsch(P, Q)`, `kabsch_umeyama(P, Q)`, `horn(P, Q)`, `horn_with_scale(P, Q)`, `kabsch_rmsd(P, Q)`, and `kabsch_umeyama_rmsd(P, Q)` signatures are identical across all frameworks. No eps parameter was ever exposed in the public API.

**3. NumPy asymmetry is expected.** NumPy's `kabsch` and `horn` do not use `finfo(dtype).eps` for RMSD clamping (they use `a_min=0.0`), while `kabsch_umeyama` and `horn_with_scale` do use `finfo(dtype).eps` for scale-denominator clamping. This is fine -- NumPy is forward-only, so the sqrt-gradient-safety concern does not apply to RMSD, and the scale denominator genuinely needs a non-zero clamp to avoid division by zero.

**4. Backward pass eps derivation is consistent.** All four autodiff frameworks now compute eps from `finfo(S.dtype).eps` (or `_DTYPE_EPS` dict for MLX) inside the backward/VJP function, using the singular-value/eigenvalue tensor's dtype. The masking logic (`abs(D) < eps`, sign-preserving replacement, diagonal protection) is structurally identical across all four.

### Verdict
APPROVED

The PR achieves its goal cleanly: the `eps` parameter is removed from internal safe wrappers in PyTorch and JAX, aligning them with TF and MLX which never had it. All frameworks now derive eps from dtype metadata consistently. The public API is unchanged. The pre-existing asymmetries noted above are all intentional framework-specific adaptations and are out of scope for this PR.

PR #138, opened 2026-03-14
fix: switch RMSD sqrt guard from max(mse, eps) to mse + eps

### LINKED ISSUE #137: Consider switching RMSD sqrt guard from max(mse, eps) to mse + eps for smoother gradients

hunter-heidenreich:
## Summary

- Replace `sqrt(max(mse, eps))` with `sqrt(mse + eps)` across all autodiff frameworks (PyTorch, JAX, TensorFlow, MLX) -- 16 sites in 8 files
- Eliminates gradient discontinuity at `mse = eps` where gradient jumped from 0 to `1/(2*sqrt(eps))`
- Standard deep learning pattern that provides smooth gradients everywhere with negligible forward-pass bias

Fixes #137

## Test plan

- [x] Full test suite passes (6916 passed, 440 skipped, 4 xfailed)
- [ ] Verify no regression in gradient-based tests (`test_gradient_verification.py`, `test_differentiability_traps.py`)

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -168,13 +168,8 @@ def horn(
 
     aligned = jnp.matmul(p, jnp.swapaxes(R, 1, 2))
     _eps = jnp.finfo(P.dtype).eps
-    rmsd = jnp.sqrt(
-        jnp.clip(
-            jnp.sum(jnp.square(aligned - q), axis=(1, 2)) / N_pts,
-            min=_eps,
-            max=None,
-        )
-    )
+    mse = jnp.sum(jnp.square(aligned - q), axis=(1, 2)) / N_pts
+    rmsd = jnp.sqrt(mse + _eps)
 
     if is_single:
         R, t, rmsd = R[0], t[0], rmsd[0]
@@ -254,9 +249,8 @@ def horn_with_scale(
         + t[:, jnp.newaxis, :]
     )
     diff = aligned_P - Q
-    rmsd = jnp.sqrt(
-        jnp.clip(jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts, min=_eps, max=None)
-    )
+    mse = jnp.sum(jnp.square(diff), axis=(1, 2)) / N_pts
+    rmsd = jnp.sqrt(mse + _eps)
 
     if is_single:
         R, t, c, rmsd = R[0], t[0], c[0], rmsd[0]
--- a/src/kabsch_horn/jax/kabsch_svd_nd.py
+++ b/src/kabsch_horn/jax/kabsch_svd_nd.py
@@ -161,8 +161,8 @@ def kabsch(
     )
 
     aligned = jnp.matmul(P, jnp.swapaxes(R, 1, 2)) + t[:, jnp.newaxis, :]
-    diff_sq = jnp.sum(jnp.square(aligned - Q), axis=(1, 2)) / N
-    rmsd = jnp.sqrt(jnp.clip(diff_sq, min=_eps, max=None))
+    mse = jnp.sum(jnp.square(aligned - Q), axis=(1, 2)) / N
+    rmsd = jnp.sqrt(mse + _eps)
 
     if is_single:
         R, t, rmsd = R[0], t[0], rmsd[0]
@@ -264,11 +264,8 @@ def kabsch_umeyama(
         c[:, jnp.newaxis, jnp.newaxis] * jnp.matmul(P, jnp.swapaxes(R, 1, 2))
         + t[:, jnp.newaxis, :]
     )
-    rmsd = jnp.sqrt(
-        jnp.clip(
-            jnp.sum(jnp.square(aligned_P - Q), axis=(1, 2)) / N, min=_eps, max=None
-        )
-    )
+    mse = jnp.sum(jnp.square(aligned_P - Q), axis=(1, 2)) / N
+    rmsd = jnp.sqrt(mse + _eps)
 
     if is_single:
         R, t, c, rmsd = R[0], t[0], c[0], rmsd[0]
--- a/src/kabsch_horn/mlx/horn_quat_3d.py
+++ b/src/kabsch_horn/mlx/horn_quat_3d.py
@@ -144,7 +144,7 @@ def horn(P: mx.array, Q: mx.array) -> tuple[mx.array, mx.array, mx.array]:
     diff = aligned - q
     mse = mx.mean(mx.sum(mx.square(diff), axis=-1), axis=-1)
     _eps = _DTYPE_EPS.get(P.dtype, 1.1920929e-7)
-    rmsd = mx.sqrt(mx.maximum(mse, _eps))
+    rmsd = mx.sqrt(mse + _eps)
 
     if orig_dtype in (mx.float16, mx.bfloat16):
         R = R.astype(orig_dtype)
@@ -264,7 +264,7 @@ def horn_with_scale(
     ) + mx.expand_dims(t, -2)
     diff = aligned_P - Q
     mse = mx.mean(mx.sum(mx.square(diff), axis=-1), axis=-1)
-    rmsd = mx.sqrt(mx.maximum(mse, _eps))
+    rmsd = mx.sqrt(mse + _eps)
 
     if orig_dtype in (mx.float16, mx.bfloat16):
         R = R.astype(orig_dtype)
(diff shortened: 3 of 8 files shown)


hunter-heidenreich:
## JAX Review

The JAX changes in this PR replace `jnp.clip(mse, min=eps)` with `mse + eps` for the RMSD square-root guard. Both patterns prevent `sqrt(0)` gradient blow-up, and the new pattern is consistent across all four call sites in the JAX subpackage. No issues found.

### Checklist

1. **`jnp.clip` vs `mse + _eps` under JIT** -- Both are JIT-safe. The `_eps` value is a Python float derived from `jnp.finfo(...).eps`, which JAX traces as a constant scalar. No tracing concern.

2. **Custom VJP functions (`safe_svd`, `safe_eigh`) unaffected** -- Confirmed. The RMSD computation sits downstream of the custom VJP calls; `_fwd`/`_bwd` for both `safe_svd` and `safe_eigh` are unchanged. The backward-pass epsilon masking logic (lines 63-68 in `kabsch_svd_nd.py`, lines 38-44 in `horn_quat_3d.py`) is not touched.

3. **Variable rename `diff_sq` -> `mse`** -- Cosmetic only (`kabsch_svd_nd.py` line 164). The name `mse` is more accurate (it is a mean squared error). No semantic change.

4. **Functional purity** -- No in-place mutations introduced. All expressions remain pure JAX operations, compatible with `jit`, `grad`, and `vmap`.

5. **Tracing correctness** -- `mse + _eps` produces a traced `jnp.ndarray` under `jit` (addition of a traced array and a Python float constant). This is standard JAX usage and traces identically across calls.

### Issues

No issues found.

### Verdict

APPROVED

hunter-heidenreich:
## PyTorch Review

The PyTorch changes in this PR replace `torch.sqrt(torch.clamp(mse, min=eps))` with `torch.sqrt(mse + eps)` across all four RMSD computation sites in `kabsch_svd_nd.py` and `horn_quat_3d.py`. The change is correct and consistent.

### Autograd graph

`mse + _eps` where `_eps` is a Python float produced by `torch.finfo(P.dtype).eps`: PyTorch's `__add__` dispatches this to `torch.add(mse, _eps)`, which creates a valid autograd node identical in cost and graph structure to the previous `torch.clamp`. The gradient `d(mse + eps)/d(mse) = 1` is simpler than `clamp`'s conditional gradient, which is a marginal win for backward clarity.

### In-place mutation

No in-place operations introduced. All four sites use out-of-place `mse + _eps` and assign to a new `mse` local or feed directly into `torch.sqrt`. Clean.

### Python float vs tensor scalar

`torch.finfo(P.dtype).eps` returns a Python `float`, not a `torch.Tensor`. Addition of a Python scalar to a tensor is a standard PyTorch op -- it does not create a 0-d tensor on a different device or interfere with device placement. No issue here.

### SafeSVD / SafeEigh backward passes

The custom backward passes in `SafeSVD` (lines 37-98 of `kabsch_svd_nd.py`) and `SafeEigh` (lines 16-43 of `horn_quat_3d.py`) are completely untouched by this PR. The RMSD computation is downstream of the SVD/Eigh forward outputs, so the backward pass through `sqrt(mse + eps)` feeds into the standard autograd chain before reaching the custom backward. No interaction.

### torch.compile / scripting compatibility

`torch.clamp` and `torch.add` (scalar) are both supported by `torch.compile` and `torch.jit.script`. The replacement does not introduce any construct that would break compilation. The intermediate `mse` local variable is trivially optimized away by both the JIT and dynamo.

### Semantic difference: clamp vs additive eps

The old code `clamp(mse, min=eps)` enforced `rmsd >= sqrt(eps)`. The new code `sqrt(mse + eps)` enforces `rmsd >= sqrt(eps)` as well (since `mse >= 0` by construction, `mse + eps >= eps`). The practical difference is that for `mse` values between 0 and `eps`:
- Old: `sqrt(eps)` (flat, zero gradient w.r.t. mse)
- New: `sqrt(mse + eps)` (non-zero gradient, smooth)

This is the stated purpose of the PR -- keeping the gradient smooth near zero MSE -- and it is the correct choice for training stability.

### Issues

#### [NITPICK] Comment style consistency
`kabsch_svd_nd.py:196` has the comment `# RMSD (mse + eps for smooth sqrt gradient near zero)` which reads well. The corresponding sites in `horn_quat_3d.py` (lines 147-150 and 272-274) have the bare `# RMSD` comment without the parenthetical explanation. Minor inconsistency, not worth blocking.

### Verdict

APPROVED

hunter-heidenreich:
## MLX Review

The MLX changes in this PR replace `mx.sqrt(mx.maximum(mse, _eps))` with `mx.sqrt(mse + _eps)` across all four RMSD call sites in `kabsch_svd_nd.py` and `horn_quat_3d.py`. This aligns the MLX backend with the same change applied to PyTorch, JAX, and TensorFlow.

I reviewed both files in full (`src/kabsch_horn/mlx/kabsch_svd_nd.py` and `src/kabsch_horn/mlx/horn_quat_3d.py`) for MLX-specific concerns.

**Correctness of the new pattern.** `mse + _eps` unconditionally shifts the sqrt argument away from zero, which is the right thing for gradient safety -- `d/dx sqrt(x)` diverges as `x -> 0`, and the additive bias keeps the argument at least `_eps` above zero regardless of the MSE value. The old `mx.maximum(mse, _eps)` pattern clamped *to* `_eps` rather than shifting *by* `_eps`, which has a flat gradient region (grad of `maximum(x, c)` is 0 when `x < c`). The new approach is strictly better for training losses because gradients always flow.

**`_eps` is a Python float; addition with `mx.array` is valid.** `_DTYPE_EPS.get(P.dtype, 1.1920929e-7)` returns a Python `float`. MLX handles `mx.array + float` by promoting the scalar, so this works correctly and does not require an explicit `mx.array(...)` wrap. No Metal backend issues with scalar addition.

**Lazy evaluation graph is equivalent.** Both `mx.maximum` and addition are single-node ops in the MLX lazy graph. Swapping one for the other does not change graph complexity or introduce any tracing concerns.

**3D restriction unaffected.** The `dim=3` guards in both `kabsch` and `kabsch_umeyama` (`P.shape[-1] != 3` check) are untouched. The hardcoded 3x3 determinant logic is unchanged.

**Consistency with `kabsch_umeyama` scale denominator.** I note that `kabsch_umeyama` (line 279) and `horn_with_scale` (line 256) still use `mx.maximum(var_P, _eps)` for the scale denominator. This is correct -- the scale denominator is a division guard, not an sqrt guard, so `maximum` (clamping) is the right choice there. The PR correctly only changes the sqrt sites.

### Issues

No blocking, recommended, or nitpick-level issues found for the MLX backend.

### Verdict

APPROVED

PR #168, opened 2026-03-15
fix: MLX SVD backward, global device leak, and NaN abort

### LINKED ISSUE #152: fix(mlx): SVD backward produces incorrect gradients for coincident singular values
## Summary

The `safe_svd` backward in the MLX backend incorrectly handles near-zero off-diagonal entries in `S_sq_diff`, producing `inf` values that are silently masked to zero. This means gradient-based optimization on symmetric or degenerate point clouds receives incorrect gradient signals -- the exact scenario the safe wrapper exists to prevent.

## Location

- `src/kabsch_horn/mlx/kabsch_svd_nd.py:51` — zero replacement
- `src/kabsch_horn/mlx/kabsch_svd_nd.py:66` — NaN mask (band-aid)

## Problem

Line 51:
```python
S_sq_diff_safe = mx.where(mx.abs(S_sq_diff) < eps, eye * eps, S_sq_diff)
```

For off-diagonal entries where `|s_i^2 - s_j^2| < eps` (coincident singular values), `eye[i,j]` is `0`, so the replacement value is `0 * eps = 0`. The subsequent `F = 1.0 / S_sq_diff_safe` produces `inf` for those entries.

The NaN mask at line 66 catches the resulting NaNs downstream:
```python
dA = mx.where(mx.isnan(dA), mx.zeros_like(dA), dA)
```

But this silently returns zeroed (incorrect) gradients rather than the correct safe gradient.

## Correct pattern

The `safe_eigh_bwd` in `horn_quat_3d.py:26` already uses the correct approach:
```python
safe_D = mx.where(mask, mx.where(D >= 0, eps, -eps), D)
```

This preserves the sign for all entries (including off-diagonal). The SVD backward should use an analogous sign-preserving replacement, then separately handle the diagonal.

## Impact

Affects any training scenario where point clouds are symmetric or have repeated singular values in the cross-covariance matrix (e.g., identity mapping, coplanar points, regular polygons).

## Found by

MLX framework audit (gradient correctness)

### LINKED ISSUE #154: fix(mlx): _warn_if_float64 permanently changes global default device to CPU
## Summary

The `_warn_if_float64` utility function calls `mx.set_default_device(mx.cpu)` as a persistent, process-wide side effect. All subsequent MLX operations -- including unrelated user code -- will run on CPU until the user manually resets the default device.

## Location

- `src/kabsch_horn/mlx/_utils.py:34`

## Problem

```python
def _warn_if_float64(P, Q, stacklevel=3):
    if P.dtype == mx.float64 or Q.dtype == mx.float64:
        warnings.warn(...)
        mx.set_default_device(mx.cpu)  # persistent global side effect
```

The warning message documents this behavior, which is good. But the function does not restore the original default device after the computation completes, so a single call with float64 inputs permanently degrades performance for all subsequent MLX operations in the process.

This also causes device-state leakage between test cases (the test adapter at `tests/adapters.py:398` triggers the same path).

## Suggested fix

Instead of changing the global default device, use `stream=mx.cpu` on individual operations that need CPU (SVD already does this). Alternatively, save and restore the default device:

```python
original_device = mx.default_device()
mx.set_default_device(mx.cpu)
try:
    # ... computation ...
finally:
    mx.set_default_device(original_device)
```

## Found by

MLX framework audit (safety)

### LINKED ISSUE #155: fix(mlx): NaN inputs to mx.linalg.svd fatally abort the process
## Summary

MLX's `mx.linalg.svd` and `mx.linalg.eigh` fatally abort the entire Python process when given NaN inputs. Neither `safe_svd` nor `safe_eigh_fwd` guard against this, making a process-killing crash the default behavior for any NaN propagation into the MLX backend.

## Locations

- `src/kabsch_horn/mlx/kabsch_svd_nd.py:13` -- `safe_svd` calls `mx.linalg.svd`
- `src/kabsch_horn/mlx/horn_quat_3d.py:8` -- `safe_eigh_fwd` calls `mx.linalg.eigh`

## Problem

Unlike PyTorch, which catches `torch.linalg.LinAlgError` and returns NaN tensors gracefully (`pytorch/kabsch_svd_nd.py:19-31`), the MLX backend has no equivalent protection. If NaN values propagate into the cross-covariance matrix H (e.g., from upstream gradient issues or bad input data), the entire process crashes with no recovery and no stack trace.

The test adapter already documents this:
```python
supports_nan_input = False  # MLX linalg.svd fatally aborts the process on NaN inputs
```

A silent process abort is the worst possible failure mode for a library. Users get no error message, no stack trace, and no chance to handle the error.

## Fix

Add a NaN guard on the small intermediate matrix immediately before the linalg call. Check the cross-covariance matrix `H` (shape `[..., D, D]`, typically 3x3) right before SVD, and the Horn matrix `N_mat` (shape `[..., 4, 4]`) right before eigh. If NaNs are detected, return NaN-filled tensors with the correct shapes, matching PyTorch's contract.

For `safe_svd`:
```python
mx.eval(A)
if mx.any(mx.isnan(A)).item():
    nan_u = mx.full((*A.shape[:-1], A.shape[-2]), float("nan"), dtype=A.dtype)
    nan_s = mx.full((*A.shape[:-2], min(A.shape[-2], A.shape[-1])), float("nan"), dtype=A.dtype)
    nan_vt = mx.full((*A.shape[:-2], A.shape[-1], A.shape[-1]), float("nan"), dtype=A.dtype)
    return nan_u, nan_s, nan_vt
```

Same pattern for `safe_eigh_fwd`.

This requires `mx.eval()` which breaks lazy evaluation, but only on the small D x D or 4 x 4 matrix immediately before the decomposition. The cost is negligible compared to the decomposition itself.

## Success criteria

- `safe_svd` and `safe_eigh_fwd` return NaN-filled tensors (correct shapes and dtype) when the input contains NaN, instead of aborting
- `supports_nan_input` in `tests/adapters.py` can be flipped to `True` for the MLX adapter
- NaN-input tests that currently skip MLX should pass
- No measurable regression on non-NaN inputs (the eval is on a tiny matrix)
- Existing tests continue to pass

## Found by

MLX framework audit (safety)

hunter-heidenreich:
## Summary

- **#152 (critical)**: `safe_svd_bwd` used `eye * eps` replacement which zeroed gradients for coincident singular values. Replaced with sign-preserving masking matching PyTorch and `safe_eigh_bwd`.
- **#154 (moderate)**: `_warn_if_float64` permanently set the global default device to CPU via `mx.set_default_device(mx.cpu)`. Replaced with a `_float64_device_guard` context manager that restores the original device after the call.
- **#155 (moderate)**: NaN inputs to `mx.linalg.svd`/`mx.linalg.eigh` fatally abort the process. Added NaN guards in `safe_svd` and `safe_eigh_fwd` that return NaN-filled tensors with correct shapes.

Closes #152, closes #154, closes #155.

## Test plan

- [x] `uv run pytest tests/ -k "mlx or MLX"` -- 821 passed, 28 skipped (non-3D dim filters)
- [x] `uv run pytest tests/` -- 4857 passed, 136 skipped, 4 xfailed (1 pre-existing TF Hypothesis flake)
- [x] NaN tests in `test_error_handling.py::test_propagates_nans_gracefully` now run for MLX instead of skipping

### DIFF
--- a/src/kabsch_horn/mlx/_utils.py
+++ b/src/kabsch_horn/mlx/_utils.py
@@ -1,4 +1,5 @@
 import warnings
+from contextlib import contextmanager
 
 import mlx.core as mx
 
@@ -13,10 +14,8 @@ _DTYPE_EPS = {
 def _warn_if_float64(P: mx.array, Q: mx.array, stacklevel: int = 3) -> None:
     """Emit a UserWarning if either P or Q is float64.
 
-    MLX does not support float64 on GPU. Passing float64 inputs sets the
-    process-wide default device to CPU (mx.set_default_device(mx.cpu)).
-    This is a persistent side effect: all subsequent MLX operations in the
-    process will also run on CPU until the default device is changed again.
+    MLX does not support float64 on GPU. When float64 inputs are detected,
+    operations are temporarily routed to CPU via _float64_device_guard.
     stacklevel=3 is correct for direct callers (kabsch, horn, etc.).
     For rmsd wrappers (kabsch_rmsd -> kabsch -> here), stacklevel points
     one level into the wrapper rather than the user call site; this is an
@@ -24,11 +23,26 @@ def _warn_if_float64(P: mx.array, Q: mx.array, stacklevel: int = 3) -> None:
     """
     if P.dtype == mx.float64 or Q.dtype == mx.float64:
         warnings.warn(
-            "MLX does not support float64 on GPU; falling back to CPU. "
-            "This sets the process-wide default device to CPU "
-            "(mx.set_default_device(mx.cpu)) and will affect all subsequent "
-            "MLX operations in this process.",
+            "MLX does not support float64 on GPU; operations will temporarily "
+            "run on CPU for this call.",
             UserWarning,
             stacklevel=stacklevel,
         )
+
+
+@contextmanager
+def _float64_device_guard(*tensors):
+    """Temporarily set the default device to CPU if any tensor is float64.
+
+    Restores the original default device on exit.
+    """
+    needs_cpu = any(t.dtype == mx.float64 for t in tensors)
+    if needs_cpu:
+        original = mx.default_device()
         mx.set_default_device(mx.cpu)
+        try:
+            yield
+        finally:
+            mx.set_default_device(original)
+    else:
+        yield
--- a/src/kabsch_horn/mlx/_utils.py
+++ b/src/kabsch_horn/mlx/_utils.py
@@ -31,10 +31,12 @@ def _warn_if_float64(P: mx.array, Q: mx.array, stacklevel: int = 3) -> None:
 
 
 @contextmanager
-def _float64_device_guard(*tensors):
+def _float64_device_guard(*tensors: mx.array):
     """Temporarily set the default device to CPU if any tensor is float64.
 
-    Restores the original default device on exit.
+    Restores the original default device on exit. Note: mx.set_default_device
+    is process-global, so this guard is not thread-safe. This is acceptable
+    for typical single-threaded MLX usage.
     """
     needs_cpu = any(t.dtype == mx.float64 for t in tensors)
     if needs_cpu:
--- a/src/kabsch_horn/mlx/horn_quat_3d.py
+++ b/src/kabsch_horn/mlx/horn_quat_3d.py
@@ -1,10 +1,16 @@
 import mlx.core as mx
 
-from ._utils import _DTYPE_EPS, _warn_if_float64
+from ._utils import _DTYPE_EPS, _float64_device_guard, _warn_if_float64
 
 
 @mx.custom_function
 def safe_eigh_fwd(A: mx.array) -> tuple[mx.array, mx.array]:
+    mx.eval(A)
+    if mx.any(mx.isnan(A)).item():
+        n = A.shape[-1]
+        nan_l = mx.full((*A.shape[:-2], n), float("nan"), dtype=A.dtype)
+        nan_v = mx.full_like(A, float("nan"))
+        return nan_l, nan_v
     L, V = mx.linalg.eigh(A, stream=mx.cpu)
     return L, V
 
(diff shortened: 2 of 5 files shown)


hunter-heidenreich:
## API Consistency Review

The MLX changes fix three issues (SVD backward, global device leak, NaN abort) while maintaining API-level consistency with other frameworks. Return types/shapes and exception types are unchanged. Two items warrant attention.

### Issues

#### [RECOMMENDED] MLX NaN guard forces eager evaluation, diverging from other frameworks' lazy behavior

`safe_svd` and `safe_eigh_fwd` now call `mx.eval(A)` followed by `mx.any(mx.isnan(A)).item()` before the decomposition (`kabsch_svd_nd.py:13-14`, `horn_quat_3d.py:8-9`). This forces eager materialization of the computation graph at the SVD boundary, which no other framework does. PyTorch catches NaN via `LinAlgError` exception handling (lazy), JAX propagates NaN through `jnp.linalg.svd` natively (lazy), and TensorFlow propagates NaN through `tf.linalg.svd` natively (lazy). The observable output is the same (NaN in, NaN out), but the eager eval could subtly affect performance or interaction with `mx.compile`/`mx.grad` in ways that differ from other frameworks. This is likely necessary as a workaround for MLX's fatal abort on NaN SVD inputs, so it is acceptable -- but consider adding a brief note to the `safe_svd` docstring explaining why `mx.eval` is needed (MLX crashes otherwise).

#### [RECOMMENDED] MLXAdapter._set_device still permanently sets device in test adapter

The production code now properly scopes the CPU device switch via `_float64_device_guard` (context manager that restores the original device). However, `MLXAdapter._set_device()` in `tests/adapters.py:437-441` still calls `mx.set_default_device(mx.cpu)` / `mx.set_default_device(mx.gpu)` permanently, and it is called from every adapter method (`kabsch`, `horn`, `convert_in`, `get_grad`, etc.). This means the test adapter still has the global side effect that the production code just fixed. While this is test-only code and not user-facing, it creates a mismatch: tests exercise a different device-management path than production. If a test calls `adapter.kabsch()` followed by non-adapter MLX code, the default device would be permanently set by the adapter. Consider updating `_set_device` to be a no-op or context manager now that the production code handles device scoping internally.

#### [NITPICK] Warning message text changed -- minor API-visible change

The `_warn_if_float64` warning text changed from `"MLX does not support float64 on GPU; falling back to CPU. This sets the process-wide default device to CPU (mx.set_default_device(mx.cpu)) and will affect all subsequent MLX operations in this process."` to `"MLX does not support float64 on GPU; operations will temporarily run on CPU for this call."` (`_utils.py:26-27`). This is technically API-visible if downstream code uses `warnings.filterwarnings` with a `message` regex match, or captures warning text in tests. The new text is more accurate given the scoped behavior, so this is the right change. Just noting it for completeness.

### Verdict
APPROVED

hunter-heidenreich:
## MLX Review

The PR fixes three real issues (SVD backward sign masking, global device leak, NaN abort) with generally correct approaches. A few items need attention, mostly around the sign discrepancy between the MLX and PyTorch SVD backward formulas.

### Issues

#### [BLOCKING] safe_svd_bwd uses additive J/K terms but PyTorch uses subtractive -- verify sign convention

In `kabsch_svd_nd.py:74`, the MLX backward computes:

```python
term = dS_mat + mx.matmul(J, S_mat) + mx.matmul(S_mat, K)
```

The PyTorch backward (`pytorch/kabsch_svd_nd.py:99`) computes:

```python
term = S_diag - torch.matmul(J, S_mat) - torch.matmul(S_mat, K)
```

This is not merely a sign-of-J/K difference -- the MLX code uses `+` for both J and K terms while PyTorch uses `-` for both. The comment on line 72-73 ("We will test +/-") suggests this was intentionally chosen, but the reason is not documented. The sign difference is justified only if J and K in MLX absorb an extra negation compared to PyTorch. Looking at how J is computed:

- PyTorch: `J = F * (Ut_dU - Ut_dU.mH)` and `K = F * (Vht_dVh - Vht_dVh.mH)` (note: K uses `Vh @ dVh.mH`, not `V^T @ dV`)
- MLX: `J = F * (Ut_dU - Ut_dU^T)` and `K = F * (Vt_dV - Vt_dV^T)` where `Vt_dV = Vt @ dVt^T`

The PyTorch K computation uses `Vht_dVh = Vh @ grad_Vh.mH = Vh @ (grad_V.mH).mH = Vh @ grad_V`. But MLX computes `Vt_dV = Vt @ dV = Vt @ dVt^T`. Since MLX's `safe_svd` returns `Vt` (not `V`), and the cotangent `dVt` comes in for `Vt`, we have `dV = dVt^T`. So `Vt_dV = Vt @ dVt^T`. In PyTorch, `Vht_dVh = Vh @ grad_V` where `Vh = Vt` and `grad_V = dV = dVt^T`. So `Vht_dVh = Vt @ dVt^T` -- identical.

Similarly for J: both compute `U^T @ dU`, identical.

So J and K are the same in both implementations, meaning the `+` vs `-` sign difference in `term` is a genuine discrepancy. One of them must be wrong. Since PyTorch's SVD backward is well-established and matches the standard formula from [Townsend 2016], the MLX version likely has the wrong sign. This could produce incorrect gradients that happen to pass tests by coincidence (e.g., if tests only exercise near-identity rotations where J and K contributions are small).

Please verify with a non-trivial gradient check that the `+` signs are correct, or switch to `-` to match PyTorch. The "We will test +/-" comment should be replaced with a justification either way.

**File**: `src/kabsch_horn/mlx/kabsch_svd_nd.py:72-76`

#### [RECOMMENDED] mx.eval() in @mx.custom_function breaks lazy graph and is incompatible with mx.compile

`safe_svd` (line 13) and `safe_eigh_fwd` (line 8 of `horn_quat_3d.py`) call `mx.eval(A)` followed by `mx.any(mx.isnan(A)).item()` to check for NaN inputs. This forces materialization of the entire computation graph up to that point, which:

1. Defeats MLX's lazy evaluation / graph fusion. Any operations upstream of the SVD/eigh call will be evaluated eagerly rather than fused into a single kernel launch.
2. Will likely break under `mx.compile()`, which traces the function and does not support data-dependent control flow (`if ... .item()`). If users wrap their training loop in `mx.compile`, this will error or produce incorrect results.

If MLX fatally aborts on NaN (the original issue), then the `mx.eval` + check approach is a reasonable pragmatic workaround, but it should be documented as incompatible with `mx.compile` (e.g., a note in the docstring).

**Files**: `src/kabsch_horn/mlx/kabsch_svd_nd.py:13-14`, `src/kabsch_horn/mlx/horn_quat_3d.py:8-9`

#### [RECOMMENDED] _float64_device_guard is not thread-safe

`mx.set_default_device()` is a process-global setting. If two threads call `kabsch()` concurrently -- one with float64 and one with float32 -- the float64 call will set the default to CPU, and the float32 call's restore in `finally` could set it back to GPU while the float64 call is still running (or vice versa). This is the same class of bug as the original device leak, just narrowed to concurrent use.

This is probably acceptable for most MLX workloads (single-threaded), but a brief doc comment noting the limitation would be helpful.

**File**: `src/kabsch_horn/mlx/_utils.py:33-48`

#### [NITPICK] stream=mx.cpu is hardcoded for SVD and eigh

`mx.linalg.svd(A, stream=mx.cpu)` on `kabsch_svd_nd.py:21` and `mx.linalg.eigh(A, stream=mx.cpu)` on `horn_quat_3d.py:14` always force CPU execution regardless of dtype. This means even float32 inputs pay the CPU-dispatch cost for the decomposition step. If MLX's Metal SVD/eigh becomes stable for float32 in the future, this will leave performance on the table. Consider gating on dtype (only force CPU for float64) or adding a TODO comment.

#### [NITPICK] Residual "We will test +/-" comment

The comment on `kabsch_svd_nd.py:72-73` reads like a development note rather than documentation:

```python
# For PyTorch and tf we found term = dS + J@S + S@K or similar.
# We will test +/-
```

This should either be removed or replaced with a proper explanation of the sign convention chosen.

### Verdict
CHANGES REQUESTED

hunter-heidenreich:
## Code Style Review

Both changed blocks pass `ruff check`, stay well within the 88-character line limit (longest changed line is 77 chars), and follow consistent formatting with each other and the surrounding code.

### Issues

No issues found.

### Verdict

APPROVED

hunter-heidenreich:
## Test Architecture Review

Flipping `supports_nan_input` to `True` for MLX is correct given the new NaN-guard in `safe_svd` and `safe_eigh` (lines 14-20 of `kabsch_svd_nd.py`), which intercepts NaN inputs before they reach `mx.linalg.svd`. The adapter change is consistent with the base class contract. However, there are gaps in test coverage for the device-leak fix and no Horn NaN coverage.

### Issues

#### [RECOMMENDED] No test verifies device restoration after float64 calls (issue #154)
The `_float64_device_guard` context manager (`src/kabsch_horn/mlx/_utils.py:34-48`) saves and restores the default device, which is the fix for #154. However, there is no test that asserts the device is restored after a float64 call. A test like "call kabsch with float64 inputs while GPU is default, then assert `mx.default_device() == mx.gpu` afterwards" would directly verify the fix and prevent regression. Without it, the guard could silently break.

#### [RECOMMENDED] Adapter `_set_device()` is now redundant and may mask device-leak regressions
Every `MLXAdapter` method calls `self._set_device()` before invoking the library (e.g., `adapters.py:479`). The library now has its own `_float64_device_guard` that handles device switching internally. The adapter's `_set_device()` calls reset the global device *before* the library runs, which means even if `_float64_device_guard` were removed or broken, the adapter would paper over the bug in tests. Consider removing the `_set_device()` calls from the adapter methods (keep only the one in `convert_in`) so tests exercise the library's guard as a real caller would.

#### [RECOMMENDED] NaN propagation test does not cover Horn algorithms for MLX
`test_propagates_nans_gracefully` (`test_error_handling.py:94`) is parametrized with `[a for a in ALGORITHMS if a not in ALGORITHMS_3D_ONLY]`, so it only tests `kabsch` and `umeyama`. The NaN guard was also added to `horn_quat_3d.py` (line 9), but there is no test exercising NaN propagation through `horn` or `horn_with_scale` for any framework. Consider adding a parallel test for the 3D-only algorithms.

#### [NITPICK] Skip message still references the old crash behavior
`test_error_handling.py:108-110` reads: *"MLX linalg.svd currently throws a fatal hardware Abort on NaNs which aborts pytest."* Now that MLX `supports_nan_input` is `True`, this skip path is no longer hit for MLX, so the stale message is harmless. But if it ever triggers again (e.g., NumPy), the message is misleading. Consider updating it to a generic message like: *"Framework does not support NaN propagation through SVD."*

### Verdict
CHANGES REQUESTED

hunter-heidenreich:
## Correctness Review

The sign-preserving SVD backward, NaN guards, and device-guard changes are numerically correct. All 88 MLX gradient verification tests pass, confirming end-to-end correctness.

### Issues

#### [RECOMMENDED] NaN guard U shape construction is fragile for non-square inputs

`kabsch_svd_nd.py:17` builds `nan_u` as `mx.full((*A.shape[:-1], M), ...)` where `M = A.shape[-2]`. Since `A.shape[:-1]` is `(..., M)`, this expands to `(..., M, M)` -- which is correct for full SVD. However, it only works because `mx.linalg.svd` returns full matrices. The PyTorch NaN guard (`pytorch/kabsch_svd_nd.py:26`) uses the more explicit `(*A.shape[:-1], A.shape[-2])` which has the same semantics but is equally implicit. Both are correct, but a comment noting "full SVD: U is [M, M]" would prevent future confusion if MLX ever introduces a `reduced` flag. This is minor since MLX Kabsch is restricted to 3x3 (square) inputs anyway.

#### [RECOMMENDED] Gradient flow through NaN guard produces zero gradients silently

When `safe_svd`'s NaN guard triggers (line 14-20), the forward returns NaN-filled tensors. The backward (`safe_svd_bwd`) will compute with NaN U, S, Vt, producing NaN in `dA`, which is then masked to zero by line 77: `dA = mx.where(mx.isnan(dA), mx.zeros_like(dA), dA)`. The same pattern exists in `safe_eigh_bwd` (line 46). This means the gradient signal is silently killed rather than propagated as NaN. This is a deliberate design choice (matching PyTorch's `grad_A` behavior) and prevents downstream NaN explosions, but users debugging a training loss plateau should be aware that NaN inputs produce zero gradients. Consider adding a brief comment at line 77 noting this contract, e.g., "NaN outputs from the forward pass produce zero gradients by design."

#### [NITPICK] Comment at line 72-73 is a leftover from development

`kabsch_svd_nd.py:72-73` says "For PyTorch and tf we found term = dS + J@S + S@K or similar. We will test +/-". This reads like a development note rather than a permanent comment. Consider replacing it with a reference to the derivation (e.g., "Standard SVD backward: dA = U (dS_diag + F_u S + S F_v) V^T, see Townsend 2016").

### Analysis Details

**Sign-preserving masking (#152 fix)**: The new code at lines 60-65 is correct. For near-zero off-diagonal entries in `S_sq_diff`, it replaces them with `+eps` or `-eps` preserving the original sign, then sets the diagonal to 1.0 before computing `1/safe_diff`, then zeros the diagonal of F. This matches the PyTorch pattern at `pytorch/kabsch_svd_nd.py:73-84` exactly. The sign preservation matters because `F[i,j]` appears in the Lorentz-covariant term `F * (skew-symmetric)`, and flipping the sign of F would flip the gradient direction for near-degenerate singular value pairs.

The MLX F matrix has opposite sign from PyTorch's (`F_mlx[i,j] = 1/(s_j^2 - s_i^2)` vs `F_pytorch[i,j] = 1/(s_i^2 - s_j^2)`) because the `expand_dims` axes are swapped. This is compensated by the MLX backward using `+J@S + S@K` vs PyTorch's `-J@S - S@K`, making the two formulas mathematically equivalent.

**NaN guard shapes**: Verified against `mx.linalg.svd` which returns full SVD. For input `[..., M, N]`: U=`[..., M, M]`, S=`[..., min(M,N)]`, Vt=`[..., N, N]`. The NaN guard shapes match. For `safe_eigh_fwd`, input `[..., n, n]` produces L=`[..., n]` and V=`[..., n, n]`; the guard uses `mx.full((*A.shape[:-2], n), ...)` and `mx.full_like(A, ...)` which are correct.

**eps values**: The `_DTYPE_EPS` values in `_utils.py` match standard machine epsilon for each dtype (float32: ~1.19e-7, float64: ~2.22e-16). These are appropriate for the sign-preserving masking and consistent with PyTorch's `torch.finfo(dtype).eps`.

### Verdict
APPROVED

PR #170, opened 2026-03-15
fix: clamp scale factor before half-precision downcast

### LINKED ISSUE #160: fix: upcast to float32 before variance division in half-precision Umeyama/Horn-with-scale
## Problem

`kabsch_umeyama` and `horn_with_scale` divide by the point cloud variance (`var_P`) to compute the scale factor `c`. In float16 (max ~65504) and bfloat16 (max ~3.39e38 but only 7 bits mantissa), this division overflows or loses catastrophic precision when inputs are near-collinear or near-origin, because the variance approaches zero.

The existing float16/bfloat16 upcast guard covers the SVD and eigendecomposition, but the variance division happens *after* the upcast-downcast boundary in some code paths, or operates on intermediate values that have already been downcast.

The README documents this as a known boundary ("variance division can overflow"), but it is mitigable.

## Proposal

Ensure the variance computation and `c = ... / var_P` division always execute in float32 (or the upcast dtype), even when the final outputs are downcast to half precision. This is a targeted fix -- the upcast scope just needs to include the variance path, not the entire function.

This relates to #89 but is more specific: #89 covers general half-precision gradient stability, while this targets the concrete overflow mechanism in scale estimation.

## Affected functions

- `kabsch_umeyama` in all five frameworks
- `horn_with_scale` in all five frameworks

hunter-heidenreich:
## Summary

Fixes #160. When `var_P` is near-zero (origin collapse, near-collinear inputs), the scale factor `c` computed in float32 can overflow to `inf` when downcast to float16/bfloat16 (max ~65504).

- Clamp `c` to `finfo(orig_dtype).max` before the downcast in all 10 source files (14 sites across NumPy, PyTorch, JAX, TensorFlow, MLX)
- Remove 2 test skips that are no longer needed (near-collinear and near-coplanar Hypothesis tests with scale + half-precision)
- Add `test_scale_factor_finite_for_half_precision_origin_collapse` verifying `c` stays finite

## Test plan

- [x] `uv run pytest tests/` (fast mode) -- 4870 passed, 176 skipped
- [x] `uv run pytest tests/test_differentiability_traps.py -k "test_scale_factor_finite" --full` -- 28 passed (all half-precision adapters)
- [x] Previously-skipped near-collinear/near-coplanar hypothesis tests now pass in `--full` mode

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -269,6 +269,7 @@ def horn_with_scale(
         if orig_dtype in (jnp.float16, jnp.bfloat16):
             R = R.astype(orig_dtype)
             t = t.astype(orig_dtype)
+            c = jnp.clip(c, max=jnp.finfo(orig_dtype).max)
             c = c.astype(orig_dtype)
             rmsd = rmsd.astype(orig_dtype)
         return R, t, c, rmsd
@@ -280,6 +281,7 @@ def horn_with_scale(
     if orig_dtype in (jnp.float16, jnp.bfloat16):
         R = R.astype(orig_dtype)
         t = t.astype(orig_dtype)
+        c = jnp.clip(c, max=jnp.finfo(orig_dtype).max)
         c = c.astype(orig_dtype)
         rmsd = rmsd.astype(orig_dtype)
     return R, t, c, rmsd
--- a/src/kabsch_horn/jax/kabsch_svd_nd.py
+++ b/src/kabsch_horn/jax/kabsch_svd_nd.py
@@ -289,6 +289,7 @@ def kabsch_umeyama(
         if orig_dtype in (jnp.float16, jnp.bfloat16):
             R = R.astype(orig_dtype)
             t = t.astype(orig_dtype)
+            c = jnp.clip(c, max=jnp.finfo(orig_dtype).max)
             c = c.astype(orig_dtype)
             rmsd = rmsd.astype(orig_dtype)
 
@@ -302,6 +303,7 @@ def kabsch_umeyama(
     if orig_dtype in (jnp.float16, jnp.bfloat16):
         R = R.astype(orig_dtype)
         t = t.astype(orig_dtype)
+        c = jnp.clip(c, max=jnp.finfo(orig_dtype).max)
         c = c.astype(orig_dtype)
         rmsd = rmsd.astype(orig_dtype)
 
--- a/src/kabsch_horn/mlx/horn_quat_3d.py
+++ b/src/kabsch_horn/mlx/horn_quat_3d.py
@@ -291,6 +291,7 @@ def horn_with_scale(
         if orig_dtype in (mx.float16, mx.bfloat16):
             R = R.astype(orig_dtype)
             t = t.astype(orig_dtype)
+            c = mx.minimum(c, mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype))
             c = c.astype(orig_dtype)
             rmsd = rmsd.astype(orig_dtype)
         return R, t, c, rmsd
--- a/src/kabsch_horn/mlx/kabsch_svd_nd.py
+++ b/src/kabsch_horn/mlx/kabsch_svd_nd.py
@@ -330,6 +330,7 @@ def kabsch_umeyama(
         if orig_dtype in (mx.float16, mx.bfloat16):
             R = R.astype(orig_dtype)
             t = t.astype(orig_dtype)
+            c = mx.minimum(c, mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype))
             c = c.astype(orig_dtype)
             rmsd = rmsd.astype(orig_dtype)
 
--- a/src/kabsch_horn/numpy/horn_quat_3d.py
+++ b/src/kabsch_horn/numpy/horn_quat_3d.py
@@ -241,6 +241,7 @@ def horn_with_scale(
     if orig_dtype in (np.float16,):
         R = R.astype(orig_dtype)
         t = t.astype(orig_dtype)
+        c = np.clip(c, a_min=None, a_max=np.finfo(orig_dtype).max)
         c = c.astype(orig_dtype)
         rmsd = rmsd.astype(orig_dtype)
     return R, t, c, rmsd
--- a/src/kabsch_horn/numpy/kabsch_svd_nd.py
+++ b/src/kabsch_horn/numpy/kabsch_svd_nd.py
@@ -225,6 +225,7 @@ def kabsch_umeyama(
     if orig_dtype in (np.float16,):
         R = R.astype(orig_dtype)
         t = t.astype(orig_dtype)
+        c = np.clip(c, a_min=None, a_max=np.finfo(orig_dtype).max)
         c = c.astype(orig_dtype)
         rmsd = rmsd.astype(orig_dtype)
     return R, t, c, rmsd
--- a/src/kabsch_horn/pytorch/horn_quat_3d.py
+++ b/src/kabsch_horn/pytorch/horn_quat_3d.py
@@ -299,6 +299,7 @@ def horn_with_scale(
         if orig_dtype in (torch.float16, torch.bfloat16):
             R = R.to(orig_dtype)
             t = t.to(orig_dtype)
+            c = torch.clamp(c, max=torch.finfo(orig_dtype).max)
             c = c.to(orig_dtype)
             rmsd = rmsd.to(orig_dtype)
         return R, t, c, rmsd
@@ -310,6 +311,7 @@ def horn_with_scale(
     if orig_dtype in (torch.float16, torch.bfloat16):
         R = R.to(orig_dtype)
         t = t.to(orig_dtype)
+        c = torch.clamp(c, max=torch.finfo(orig_dtype).max)
         c = c.to(orig_dtype)
         rmsd = rmsd.to(orig_dtype)
     return R, t, c, rmsd
--- a/src/kabsch_horn/pytorch/kabsch_svd_nd.py
+++ b/src/kabsch_horn/pytorch/kabsch_svd_nd.py
@@ -340,6 +340,7 @@ def kabsch_umeyama(
         if orig_dtype in (torch.float16, torch.bfloat16):
             R = R.to(orig_dtype)
             t = t.to(orig_dtype)
+            c = torch.clamp(c, max=torch.finfo(orig_dtype).max)
             c = c.to(orig_dtype)
             rmsd = rmsd.to(orig_dtype)
         return R, t, c, rmsd
@@ -352,6 +353,7 @@ def kabsch_umeyama(
     if orig_dtype in (torch.float16, torch.bfloat16):
         R = R.to(orig_dtype)
         t = t.to(orig_dtype)
+        c = torch.clamp(c, max=torch.finfo(orig_dtype).max)
         c = c.to(orig_dtype)
         rmsd = rmsd.to(orig_dtype)
 
--- a/src/kabsch_horn/tensorflow/horn_quat_3d.py
+++ b/src/kabsch_horn/tensorflow/horn_quat_3d.py
@@ -310,6 +310,8 @@ def horn_with_scale(
     if orig_dtype in (tf.float16, tf.bfloat16):
         R = tf.cast(R, orig_dtype)
         t = tf.cast(t, orig_dtype)
+        c_max = tf.constant(tf.dtypes.as_dtype(orig_dtype).max, dtype=c.dtype)
+        c = tf.minimum(c, c_max)
         c = tf.cast(c, orig_dtype)
         rmsd = tf.cast(rmsd, orig_dtype)
     return R, t, c, rmsd
--- a/src/kabsch_horn/tensorflow/kabsch_svd_nd.py
+++ b/src/kabsch_horn/tensorflow/kabsch_svd_nd.py
@@ -285,6 +285,8 @@ def kabsch_umeyama(
     if orig_dtype in (tf.float16, tf.bfloat16):
         R = tf.cast(R, orig_dtype)
         t = tf.cast(t, orig_dtype)
+        c_max = tf.constant(tf.dtypes.as_dtype(orig_dtype).max, dtype=c.dtype)
+        c = tf.minimum(c, c_max)
         c = tf.cast(c, orig_dtype)
         rmsd = tf.cast(rmsd, orig_dtype)
 
--- a/tests/test_differentiability_traps.py
+++ b/tests/test_differentiability_traps.py
@@ -210,8 +210,8 @@ class TestDifferentiabilityTraps:
             adapter, "precision", "float64"
         ) in ("float16", "bfloat16"):
             pytest.skip(
-                "Scale algorithms require division by variance, which overflows "
-                "float16 on origin collapse."
+                "Scale algorithms produce large gradients on origin collapse "
+                "that overflow float16/bfloat16 range."
             )
 
         rng = np.random.default_rng(42)
@@ -306,13 +306,6 @@ class TestDifferentiabilityTraps:
         algo: str,
     ) -> None:
         """Gradients remain finite for near-collinear point clouds (Hypothesis)."""
-        if algo in ALGORITHMS_WITH_SCALE and getattr(
-            adapter, "precision", "float64"
-        ) in ("float16", "bfloat16"):
-            pytest.skip(
-                "Scale algorithms require division by variance, which can overflow "
-                "float16 on near-collinear inputs."
-            )
 
         @settings(
             max_examples=_MAX_EXAMPLES,
@@ -381,12 +374,6 @@ class TestDifferentiabilityTraps:
         algo: str,
     ) -> None:
         """Gradients remain finite for near-coplanar point clouds (Hypothesis)."""
-        if algo in ALGORITHMS_WITH_SCALE and getattr(
-            adapter, "precision", "float64"
-        ) in ("float16", "bfloat16"):
-            pytest.skip(
-                "Umeyama variance division overflows float16 on near-coplanar inputs."
-            )
         if adapter.name == "MLXAdapter" and adapter.precision == "float32":
             pytest.skip(
                 "MLX float32 SafeSVD backward is unstable for near-coplanar inputs"
@@ -450,3 +437,28 @@ class TestDifferentiabilityTraps:
             assert np.all(np.isfinite(grad))
 
         _inner()
+
+    @pytest.mark.parametrize("algo", list(ALGORITHMS_WITH_SCALE))
+    @pytest.mark.parametrize("adapter", frameworks)
+    def test_scale_factor_finite_for_half_precision_origin_collapse(
+        self,
+        adapter: FrameworkAdapter,
+        algo: str,
+        dim: int,
+    ) -> None:
+        """Scale factor c stays finite (not inf) in half-precision origin collapse."""
+        if getattr(adapter, "precision", "float64") not in ("float16", "bfloat16"):
+            pytest.skip("Only relevant for half-precision adapters.")
+
+        rng = np.random.default_rng(42)
+        P_np = np.zeros((5, dim), dtype=np.float64)
+        Q_np = rng.random((5, dim)).astype(np.float64)
+
+        P = adapter.convert_in(P_np)
+        Q = adapter.convert_in(Q_np)
+        func = adapter.get_transform_func(algo)
+
+        res = func(P, Q)
+        c = adapter.convert_out(res[2])
+
+        assert np.all(np.isfinite(c)), f"Scale factor overflowed: {c}"
--- a/tests/test_differentiability_traps.py
+++ b/tests/test_differentiability_traps.py
@@ -440,7 +440,7 @@ class TestDifferentiabilityTraps:
 
     @pytest.mark.parametrize("algo", list(ALGORITHMS_WITH_SCALE))
     @pytest.mark.parametrize("adapter", frameworks)
-    def test_scale_factor_finite_for_half_precision_origin_collapse(
+    def test_scale_factor_finite_float16_origin_collapse(
         self,
         adapter: FrameworkAdapter,
         algo: str,


hunter-heidenreich:
## NumPy Review

The NumPy changes in this PR add `c = np.clip(c, a_min=None, a_max=np.finfo(orig_dtype).max)` before `c = c.astype(orig_dtype)` in both `kabsch_svd_nd.py` (line 228) and `horn_quat_3d.py` (line 244). This prevents `inf` when a float32 scale factor exceeds float16's max (~65504) during downcast.

I verified all four review points:

1. **`np.clip` with `a_min=None`** -- Valid API usage. NumPy documents that `a_min` and `a_max` can each be `None` to skip that bound. Confirmed at runtime: `np.clip(arr, a_min=None, a_max=50.0)` works correctly.

2. **`np.finfo(orig_dtype).max` for float16** -- Returns `65500.0` (the largest finite float16 value). This is the correct ceiling to prevent overflow on downcast. The value is computed from `orig_dtype` which is `np.float16` at the point of use, so it always matches the target type.

3. **Shape and dtype preservation** -- `np.clip` preserves the input array's shape and dtype. Since `c` is float32 (or float64) at this point and `np.finfo(np.float16).max` is a Python float, the result stays float32/float64 with the same shape. The subsequent `.astype(orig_dtype)` then safely downcasts without overflow. Confirmed at runtime.

4. **`(np.float16,)` check without bfloat16** -- Correct. NumPy has no `bfloat16` dtype (`hasattr(np, 'bfloat16')` is `False`). The tuple `(np.float16,)` covers the only low-precision NumPy float type that could overflow. Other frameworks (PyTorch, JAX, TF) handle bfloat16 separately in their own implementations.

Additional observation: only `c` (the scale factor) needs this protection. `R` is a rotation matrix with elements in [-1, 1], `t` is bounded by centroid magnitude (unlikely to exceed 65504 in practice), and `rmsd` is bounded by data range. The scale `c = trace(S*D) / var_P` is the only quantity that can blow up when `var_P` approaches zero, making it the correct and sufficient target for clamping.

### Issues

No issues found. The implementation is minimal, correct, and well-targeted.

### Verdict

APPROVED

hunter-heidenreich:
## PyTorch Review

The PR adds `torch.clamp(c, max=torch.finfo(orig_dtype).max)` before the `.to(orig_dtype)` downcast in both `kabsch_umeyama` and `horn_with_scale`, across both the single and batched return paths. I verified all five review criteria.

### Analysis

**1. `torch.clamp` with only `max=` (no `min`)**: Correct. The scale factor `c` is computed as `sum(S * S_corr) / var_P`. Singular values `S` are non-negative, `S_corr` entries are +/-1, and `var_P` is clamped to `min=eps` upstream. The result can be large and positive (when `var_P` is small relative to `Q`'s spread), which is the overflow direction. Negative scale values would require pathological input and would still be within representable range (symmetric bounds). Only the upper bound matters here.

**2. `torch.finfo(orig_dtype).max` correctness**: Verified empirically. `torch.finfo(torch.float16).max` returns 65504.0 and `torch.finfo(torch.bfloat16).max` returns ~3.39e+38. These are the correct saturation thresholds before downcast.

**3. Placement in autograd graph**: Correct. The clamp runs on the float32 `c` tensor *before* the `.to(orig_dtype)` cast. This means:
   - Forward: `c` is clamped while still in float32 precision, then safely downcast.
   - Backward: `torch.clamp` backward passes gradient of 1 when within bounds and 0 when clamped. This is well-defined and the expected behavior -- if `c` exceeds the target dtype's max, the gradient should be zero (saturated).

**4. Both paths covered**: Yes. All four sites are patched:
   - `kabsch_svd_nd.py` line 343 (single path) and line 356 (batched path)
   - `horn_quat_3d.py` line 302 (single path) and line 314 (batched path)

**5. Gradient flow**: `torch.clamp` has a well-defined backward in PyTorch: gradient passes through (multiplied by 1) when the value is within bounds, and is zeroed when clamped. This is correct behavior -- a scale factor that overflows the target dtype is already in a regime where gradients are meaningless. No impact on normal-range inputs.

**Why only `c`?** The other outputs (`R`, `t`, `rmsd`) are naturally bounded: `R` entries lie in [-1, 1] (rotation matrix), `t` is bounded by centroid differences (same magnitude as input coordinates), and `rmsd` is bounded by the input point spread. Scale `c` is the only output that can realistically overflow float16's 65504 max, since `c = trace / var_P` and small `var_P` amplifies the numerator.

### Issues

No issues found.

### Verdict

APPROVED

hunter-heidenreich:
## JAX Review

The PR adds `jnp.clip(c, max=jnp.finfo(orig_dtype).max)` before downcasting the scale factor `c` from float32 to float16/bfloat16. This prevents overflow to `inf` when `c` (computed in float32) exceeds the target dtype's max representable value (65504 for float16, ~3.39e38 for bfloat16).

I verified all five review points:

**1. `jnp.clip` with only `max=` kwarg** -- Valid JAX API. `jnp.clip(a, a_min=None, a_max=...)` is the documented signature; passing `max=` as a keyword alias works and leaves the lower bound unconstrained (no-op). This has been stable since JAX 0.2.x.

**2. `jnp.finfo(orig_dtype).max` for float16/bfloat16** -- Returns the correct maximum finite value for the dtype. `jnp.finfo` delegates to `numpy.finfo` for standard dtypes and handles bfloat16 via JAX's own dtype registration. `jnp.finfo(jnp.float16).max` = 65504.0, `jnp.finfo(jnp.bfloat16).max` ~= 3.389e38. Both are correct.

**3. JIT compatibility** -- `jnp.clip` is a pure functional operation with no side effects and is fully traceable by JAX's JIT compiler. The `jnp.finfo(orig_dtype).max` call resolves to a Python float constant at trace time (since `orig_dtype` is a concrete value known before tracing), so it becomes a static bound in the compiled XLA computation. No issues with `jit`, `vmap`, or `grad`.

**4. Both single and batched paths covered** -- Confirmed. In `kabsch_svd_nd.py`: line 292 (single, `is_single` branch) and line 306 (batched, post-reshape branch). In `horn_quat_3d.py`: line 272 (single) and line 284 (batched). All four insertion sites are present and correctly placed immediately before the `.astype(orig_dtype)` call.

**5. Custom VJP scope** -- The clip is applied well after `safe_svd` / `safe_eigh` have returned their results. It operates on the final scalar `c` value, outside any `custom_vjp`-decorated function. No interaction with the custom backward passes. For the forward-only path (no gradient), this is straightforwardly correct. For differentiable use, `jnp.clip` has well-defined gradients (pass-through when not clipping, zero when clipping at the bound), which is the desired behavior -- if `c` is saturated at the dtype max, the gradient correctly becomes zero.

### Issues

#### [NITPICK] Blank line left behind after skip removal

In `test_differentiability_traps.py`, removing the `if algo in ALGORITHMS_WITH_SCALE...` skip block in `test_gradients_stable_nearly_collinear_hypothesis` leaves a blank line between the docstring and the `@settings` decorator (visible in the diff at the line after the docstring). This is cosmetic and non-blocking.

### Verdict

APPROVED

hunter-heidenreich:
## API Consistency Review

All 10 source files (`kabsch_svd_nd.py` and `horn_quat_3d.py` for NumPy, PyTorch, JAX, TensorFlow, MLX) were read and compared for consistent application of the scale-factor clamp before downcast.

### Scope of the clamp

The clamp applies only to `c` (the Umeyama/Horn scale factor) in `kabsch_umeyama` and `horn_with_scale`. The `kabsch` and `horn` functions do not return a scale factor, so they correctly have no clamp -- verified across all 5 frameworks.

### Clamp pattern by framework

| Framework   | Function            | Clamp idiom                                                        | Blocks | Correct position? |
|-------------|---------------------|--------------------------------------------------------------------|--------|--------------------|
| NumPy       | `kabsch_umeyama`    | `np.clip(c, a_min=None, a_max=np.finfo(orig_dtype).max)`          | 1      | Yes (after reshape, before astype) |
| NumPy       | `horn_with_scale`   | `np.clip(c, a_min=None, a_max=np.finfo(orig_dtype).max)`          | 1      | Yes |
| PyTorch     | `kabsch_umeyama`    | `torch.clamp(c, max=torch.finfo(orig_dtype).max)`                 | 2      | Yes (single + batched) |
| PyTorch     | `horn_with_scale`   | `torch.clamp(c, max=torch.finfo(orig_dtype).max)`                 | 2      | Yes (single + batched) |
| JAX         | `kabsch_umeyama`    | `jnp.clip(c, max=jnp.finfo(orig_dtype).max)`                      | 2      | Yes (single + batched) |
| JAX         | `horn_with_scale`   | `jnp.clip(c, max=jnp.finfo(orig_dtype).max)`                      | 2      | Yes (single + batched) |
| TensorFlow  | `kabsch_umeyama`    | `c_max = tf.constant(...); c = tf.minimum(c, c_max)` | 1 | Yes |
| TensorFlow  | `horn_with_scale`   | `c_max = tf.constant(...); c = tf.minimum(c, c_max)` | 1 | Yes |
| MLX         | `kabsch_umeyama`    | `mx.minimum(c, mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype))` | 1     | Yes |
| MLX         | `horn_with_scale`   | `mx.minimum(c, mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype))` | 1     | Yes |

### Detailed findings

**1. Semantic equivalence -- PASS.** All 10 sites clamp `c` to `finfo(orig_dtype).max` (upper bound only). The clamp is one-sided (no lower bound), which is correct: `c` is always positive by construction (trace(S * D) / var_P with both terms non-negative). Each framework uses its native clamp/clip/minimum primitive, and the bound value is always `finfo(orig_dtype).max` for the *original* (pre-upcast) dtype.

**2. Control-flow placement -- PASS.** In every case the clamp is applied:
- After `c` has been reshaped/indexed to its final shape (post-`view`/`reshape`/`[0]`)
- Before the downcast to `orig_dtype` (`astype`/`.to()`/`tf.cast`)
- Inside the `if orig_dtype in (float16, bfloat16)` guard

PyTorch and JAX each have two downcast blocks (single-input early return + batched path). Both blocks include the clamp in all four affected functions. NumPy, TF, and MLX have a single downcast block each, all covered.

**3. Coverage -- PASS.** Every `kabsch_umeyama` and `horn_with_scale` function across all 5 frameworks has the clamp. Every downcast site within those functions is covered:
- PyTorch `kabsch_umeyama`: lines 343 and 356
- PyTorch `horn_with_scale`: lines 302 and 314
- JAX `kabsch_umeyama`: lines 292 and 306
- JAX `horn_with_scale`: lines 272 and 284
- NumPy `kabsch_umeyama`: line 228
- NumPy `horn_with_scale`: line 244
- TF `kabsch_umeyama`: lines 288-289
- TF `horn_with_scale`: lines 313-314
- MLX `kabsch_umeyama`: line 333
- MLX `horn_with_scale`: line 294

**4. Return shapes/types -- PASS.** The clamp is scalar-to-scalar (or batch-to-batch) and preserves the shape of `c`. The subsequent `astype`/`.to()`/`tf.cast` preserves the dtype contract. Return signatures are unchanged: `(R, t, c, rmsd)` with the documented shapes.

### Issues

#### [NITPICK] TensorFlow clamp idiom differs from the other frameworks

TensorFlow uses a two-step pattern (`tf.constant` + `tf.minimum`) while all other frameworks use a single call (`clip`/`clamp`/`minimum` with an inline max value). This is functionally correct -- TF lacks a `clamp` with a `max` keyword -- so this is just a style observation, not a correctness concern.

#### [NITPICK] NumPy only guards `float16`, others guard `float16` and `bfloat16`

NumPy uses `if orig_dtype in (np.float16,)` (single-element tuple), while PyTorch/JAX/TF/MLX check for both `float16` and `bfloat16`. This is correct because NumPy does not have a `bfloat16` dtype, so the guard is appropriately narrower. No action needed.

### Verdict

**APPROVED** -- The clamp pattern is semantically identical, correctly placed, and covers all downcast sites across all 10 files. Return shapes and types are unchanged.

hunter-heidenreich:
## MLX Review

The two MLX files each add a single line clamping the scale factor `c` before downcasting from float32 to float16/bfloat16. This prevents `inf` values when the float32 scale exceeds the representable range of the target half-precision dtype (~65504 for float16, ~3.39e38 for bfloat16).

### Verification

#### 1. `mx.finfo(orig_dtype).max` -- valid MLX API

`mx.finfo` mirrors NumPy's `numpy.finfo` and is documented in the MLX API. It accepts an `mx.Dtype` and returns an object with `.max`, `.min`, `.eps`, etc. Using it with `mx.float16` or `mx.bfloat16` (the only dtypes that reach this branch) is correct.

#### 2. `mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype)` -- correct scalar construction

`mx.finfo(...).max` returns a Python float. Wrapping it in `mx.array(..., dtype=c.dtype)` creates a 0-d MLX array in the compute dtype (float32), which is the right dtype for the comparison against `c`. This avoids any implicit promotion or dtype mismatch in `mx.minimum`.

#### 3. `mx.minimum` broadcasting -- correct for scalar vs batched `c`

`mx.minimum` follows NumPy broadcasting semantics. A 0-d array (the bound) broadcasts against any shape of `c` (scalar, 1-d batch, or arbitrary leading batch dims). This is correct.

#### 4. MLX lazy evaluation -- no ordering issues

All three operations (`mx.finfo`, `mx.array`, `mx.minimum`) produce lazy graph nodes that will be evaluated together when `mx.eval` is eventually called (or when `.item()` is used downstream). There is no data-dependent branching or `.item()` call here, so lazy evaluation order is fine. The clamp is a pure element-wise op with no side effects.

#### 5. Single downcast block per file -- confirmed

Both MLX files (`kabsch_svd_nd.py` and `horn_quat_3d.py`) have exactly **one** downcast block per function that returns a scale factor:

- `kabsch_svd_nd.py` line 330: single `if orig_dtype in (mx.float16, mx.bfloat16):` block in `kabsch_umeyama`
- `horn_quat_3d.py` line 291: single `if orig_dtype in (mx.float16, mx.bfloat16):` block in `horn_with_scale`

This is simpler than PyTorch and JAX, which each have two downcast blocks per function (one inside a batched branch, one outside). The PR correctly adds the clamp in the single block for each MLX function -- no missing sites.

### Issues

#### [NITPICK] Comment explaining the clamp

The PyTorch and JAX changes are self-documenting (`torch.clamp`, `jnp.clip` with `max=`), but `mx.minimum(c, mx.array(mx.finfo(orig_dtype).max, dtype=c.dtype))` is more verbose and less immediately obvious. A short inline comment like `# clamp to prevent inf on downcast` would help readability, matching the style of other inline comments in these files (e.g., the `# Correction` comment near the determinant sign fix). This is purely cosmetic.

### Verdict

APPROVED -- the MLX changes are correct. Both clamp sites are covered, the API usage is valid, broadcasting works as expected, and lazy evaluation poses no issues.

hunter-heidenreich:
## TensorFlow Review

Replaces `tf.ones_like(d_sign)` + `tf.repeat(tf.expand_dims(...), dim-1)` with direct `tf.ones(tf.concat([tf.shape(d_sign), [dim - 1]], 0))` in both `kabsch` and `kabsch_umeyama`. Correct and `@tf.function`-compatible.

### Issues

No issues found. Verified:

- **Dynamic shape compatibility**: `tf.ones(tf.concat([tf.shape(d_sign), [dim - 1]], 0), dtype=d_sign.dtype)` correctly uses `tf.shape()` (dynamic) for batch dims and `[dim - 1]` (also dynamic via `tf.shape(P)[-1]`) for the trailing dimension. This traces correctly under `@tf.function` with dynamic shapes -- no Python-level shape dependency.
- **Replaces `tf.repeat`**: The old pattern `tf.repeat(tf.expand_dims(ones, -1), dim - 1, axis=-1)` first created a scalar-per-batch `ones`, expanded it, then repeated `dim-1` times. The new `tf.ones(shape)` produces the same `[..., dim-1]` tensor in one op.
- **No eager/graph divergence**: Both `tf.ones` and `tf.concat` behave identically in eager and graph mode. No `.numpy()` calls or Python control flow on tensor values introduced.
- **GradientTape**: `tf.ones` creates a constant tensor (zero gradient). `d_sign` from `tf.sign` also has zero gradient. The `diag_vals` concat is used in `tf.linalg.diag` then `tf.matmul`, where gradients flow through V and U, not through the diagonal constants. No change to gradient behavior.
- **dtype correctness**: `dtype=d_sign.dtype` preserves the dtype from the (possibly upcast) input, matching the old `tf.ones_like(d_sign)` behavior.

### Verdict
APPROVED

hunter-heidenreich:
## Tests Review

The test file consolidates `TestDifferentiabilityTraps` and the former `TestHornDifferentiabilityTraps` into a single unified class, parametrized over `ALGORITHMS` (kabsch, umeyama, horn, horn_with_scale). The new test and skip changes look correct.

### Issues

#### [NITPICK] New test follows AAA pattern cleanly

`test_scale_factor_finite_for_half_precision_origin_collapse` (lines 441-464) follows AAA well:
- **Arrange**: skip non-half-precision adapters, create origin-collapsed P and random Q
- **Act**: call `func(P, Q)` and extract scale factor `c`
- **Assert**: `np.all(np.isfinite(c))` with an informative f-string message

No issues here.

#### [NITPICK] Half-precision skip is correctly scoped

Line 450-451 skips adapters whose precision is not `float16` or `bfloat16`, which is correct -- the test specifically validates the clamp-before-downcast fix that only matters for half-precision types. The `getattr` fallback to `"float64"` handles adapters without a `precision` attribute safely.

#### [NITPICK] Scale algorithm coverage via ALGORITHMS_WITH_SCALE

Line 441 parametrizes over `ALGORITHMS_WITH_SCALE` which is `{"umeyama", "horn_with_scale"}` (conftest.py line 10). This correctly covers both SVD-based and quaternion-based scale paths, which are the only algorithms that compute `c = ... / var_P`.

#### [NITPICK] Removed `filter_too_much` health check suppression is safe

The old Hypothesis tests used `assume(adapter.supports_dim(...))` to filter generated examples, requiring `HealthCheck.filter_too_much` suppression. The new code draws dimensions from `adapter.supported_dims()` directly via `st.data()`, so filtering is no longer needed. This is a strict improvement.

#### [NITPICK] Origin collapse skip (lines 209-215) is correctly retained

The skip for `ALGORITHMS_WITH_SCALE` + half-precision on origin collapse **gradients** is still necessary. The clamp fix prevents `c` from being `inf` in the forward pass, but when `var_P` is near-zero, the *gradient* of `c = sum / var_P` involves `1 / var_P^2`, which overflows float16 range (~65504) regardless of the forward clamp. The updated skip message ("Scale algorithms produce large gradients on origin collapse that overflow float16/bfloat16 range") accurately explains this. The scope was correctly broadened from `algo == "umeyama"` to `algo in ALGORITHMS_WITH_SCALE` to also cover `horn_with_scale`.

### Verdict

APPROVED -- The new test is well-structured, the skip logic is correctly scoped, and the consolidation of Horn tests into the unified parametrized class reduces duplication without losing coverage.

hunter-heidenreich:
## Pytest Review

The test changes are well-structured overall: the skip removals are justified by the source-level clamp fix, the new test targets the exact scenario that was previously overflowing, and the `list(ALGORITHMS_WITH_SCALE)` call is correct (it is a `set`, and `parametrize` needs deterministic ordering). One issue with the collection hook interaction.

### Issues

#### [BLOCKING] New test never meaningfully executes in fast mode (without `--full`)

The new test `test_scale_factor_finite_for_half_precision_origin_collapse` is designed exclusively for float16/bfloat16 adapters (it `pytest.skip()`s all others). However, the `pytest_collection_modifyitems` hook in `conftest.py` (lines 207-218) filters out float16/bfloat16 adapter variants in fast mode unless the test name (before the `[` bracket) contains one of: `"preserves_input_dtype"`, `"preserves_dtype"`, `"float16"`, or `"dtype"`.

The test name contains `"half_precision"` -- none of those substrings match. So in fast mode:
- float16/bfloat16 adapters are removed by the collection hook
- float32/float64 adapters remain but immediately hit `pytest.skip()`
- The test never actually asserts anything

**Fix options** (pick one):
1. Rename the test to include `"float16"` in the method name, e.g. `test_scale_factor_finite_float16_origin_collapse`
2. Add `"half_precision"` to the whitelist in the collection hook filter

Option 1 is simpler and follows the existing naming convention the hook relies on.

#### [NITPICK] Blank line after skip-block removal

Line 308 (after the docstring, before `@settings`) has a trailing blank line left from the removed skip block in `test_gradients_stable_nearly_collinear_hypothesis`. This is cosmetically fine -- the existing `test_gradients_stable_nearly_coplanar_hypothesis` (line 376) does not have an equivalent blank line before its first guard, so it is a minor inconsistency, but not worth blocking over.

### Verdict

CHANGES REQUESTED -- the collection hook interaction means the new test is dead code in `uv run pytest tests/` (the default developer workflow). The fix is a one-line rename.

hunter-heidenreich:
## Correctness Review

The PR clamps the Umeyama/Horn scale factor `c` to `finfo(orig_dtype).max` before downcasting from float32 to float16/bfloat16 across all 14 downcast sites in 10 source files. This is the correct fix for the overflow-to-inf problem that occurs when `var_P` is near-zero (origin collapse, near-collinear inputs) and `c` exceeds the ~65504 representable range of float16.

**Algorithm analysis**: In both Kabsch/Umeyama (`c = sum(S * S_corr) / var_P`) and Horn (`c = trace(R^T H) / (var_P * N)`), the scale factor `c` is always non-negative by construction:
- Kabsch: singular values are non-negative and sorted descending, so `S[0] + ... + d_sign * S[-1] >= 0` (the last SV is smallest)
- Horn: `trace(R^T H)` is maximized by the optimal rotation from the eigenvector, so it is non-negative

Therefore an upper-only clamp is correct -- no lower clamp is needed.

**Site coverage**: All 14 downcast sites verified:
- JAX: 4 sites (kabsch_svd_nd: 2, horn_quat_3d: 2) -- single + batched paths
- PyTorch: 4 sites (kabsch_svd_nd: 2, horn_quat_3d: 2) -- single + batched paths
- MLX: 2 sites (1 each) -- no single/batched split
- TF: 2 sites (1 each) -- no single/batched split for downcast
- NumPy: 2 sites (1 each) -- no single/batched split for downcast

**Clamp placement**: Every clamp is correctly placed immediately before the `astype`/`to`/`cast` call and after the final computation of `c`. No computation uses the clamped value.

### Issues

#### [NITPICK] Missing lower clamp for symmetry / future-proofing

While `c >= 0` by construction in all current algorithms, the clamp is one-sided (`max` only). If someone later introduces a code path where `c` could be large-negative (unlikely given the math, but conceivable with numerical noise), the downcast would still overflow to `-inf`. Using a symmetric clamp like `clip(c, -finfo.max, finfo.max)` would be strictly safer at zero cost. This is purely defensive and not blocking.

#### [NITPICK] Removed skips may expose flaky Hypothesis tests

The two removed skips (near-collinear and near-coplanar hypothesis with scale + half-precision) are correct to remove since the forward pass now produces finite `c`. However, the *gradient* path through a clamped `c` could still overflow in half-precision for extreme inputs generated by Hypothesis. The origin-collapse gradient skip was kept (with reworded message), which is good. If CI shows flaky failures in these hypothesis tests, consider adding a forward-only variant or tightening the Hypothesis search space for half-precision.

### Verdict

APPROVED -- the fix is algorithmically correct, complete across all 14 sites, and properly placed. The upper-only clamp is justified by the non-negativity of `c` in both SVD and quaternion formulations. The test addition is well-targeted.

hunter-heidenreich:
## Hypothesis Review

The two removed `pytest.skip` blocks previously guarded `test_gradients_stable_nearly_collinear_hypothesis` and `test_gradients_stable_nearly_coplanar_hypothesis` from running with `ALGORITHMS_WITH_SCALE` (`umeyama`, `horn_with_scale`) under float16/bfloat16. The source-level clamp of `c` to `finfo(orig_dtype).max` before downcast makes these skips unnecessary.

### Analysis

**1. Safety of removing the skips**

Both tests set `Q = P.copy()`, so the scale factor `c = var_Q / var_P` is always ~1.0 regardless of input geometry. The original overflow risk was `c` exceeding float16 max (65504) when `var_P` is near-zero. With Q = P, variance cancels perfectly, so these tests were arguably safe even before the clamp fix. With the clamp in place, the safety net is explicit and the skips are correctly removed.

The origin-collapse test (`test_gradients_are_stable_when_points_collapse_to_origin`) retains its skip for scale + half-precision, which is correct -- that test creates genuinely zero-variance inputs where the *gradients* overflow, a problem the `c`-clamp alone does not solve.

**2. Strategy coverage of the overflow path**

- `nearly_collinear_3d`: Points spread along a direction with `|t_max - t_min| >= 1.0`, with perpendicular noise at 1e-4 to 1e-2 scale. Variance along the main axis is non-trivial. This would not trigger `c` overflow with Q = P, but if Q != P with different noise, the cross-covariance could produce large scale factors. Since Q = P here, no overflow risk.
- `nearly_coplanar_nd`: Main dimensions in [-10, 10], last dimension scaled to [1e-4, 1e-2]. Variance is well-bounded. Same Q = P argument applies.

Both strategies could theoretically produce large `c` values if P and Q were drawn independently with very different variances, but that scenario is not tested by these two specific tests.

**3. Hypothesis settings**

`max_examples=15` (fast) / `60` (full), `suppress_health_check=[HealthCheck.too_slow]`, `deadline=None` -- all consistent with the other Hypothesis tests in this file and appropriate for framework-backed gradient computations that have variable latency.

**4. Hypothesis-specific issues with newly exposed paths**

No issues found. The inner test functions are stateless closures that capture the adapter/algo from the outer parametrize. The `@given` decorator is applied inside the test method (nested function pattern), which is an established pattern already used by 4 other tests in this file. No database pollution or state leakage concerns.

### Issues

#### [NITPICK] Trailing blank line after skip removal

Line 309 (after the docstring in `test_gradients_stable_nearly_collinear_hypothesis`) has an extra blank line where the skip block used to be. Not a functional issue, but leaves a minor formatting artifact.

### Verdict

APPROVED -- The skip removals are safe. The Hypothesis strategies generate inputs with well-bounded variance, Q = P ensures c ~ 1.0, and the source-level clamp provides an additional safety net. Settings are consistent and appropriate.

hunter-heidenreich:
## Code Style Review

All 11 modified files pass `ruff check` with the project's line-length=88 / py310 config. No em dashes, no negation/contrastive reframes, and all lines are within the 88-character limit (longest added line is 86 chars). Writing style in skip messages and docstrings is clean and concise.

### Issues

#### [NITPICK] TF uses `c_max` temporary vs. inline in other frameworks

TensorFlow's clamp pattern introduces a named local `c_max`, while NumPy/PyTorch/JAX/MLX all inline the max value. This is justified -- TF lacks a `clip(max=...)` keyword API, so the two-line `tf.constant` + `tf.minimum` approach is idiomatic. No change needed, but worth noting for future readers that this is intentional, not an oversight.

#### [NITPICK] NumPy `np.clip` uses positional-style keyword `a_min=None`

The NumPy lines use `np.clip(c, a_min=None, a_max=...)`. While correct, the `a_min=None` is only needed because `np.clip` requires both bounds. This is idiomatic NumPy, so no change needed -- just noting the asymmetry with the other frameworks' cleaner `clip(max=...)` signatures.

### Verdict

APPROVED -- no blocking or recommended issues. The clamp patterns are idiomatic for each framework, line lengths are clean, skip messages follow the project's writing style, and the new test is well-structured.

PR #171, opened 2026-03-15
fix: reject 1D inputs with ValueError instead of IndexError

### LINKED ISSUE #153: fix(numpy): 1D array input raises IndexError instead of ValueError
## Summary

All public functions across all five frameworks access `P.shape[-2]` before verifying the input has at least 2 dimensions. A 1D input (e.g., `np.array([1.0, 2.0, 3.0])`) raises an opaque `IndexError` (or framework equivalent) instead of a clear `ValueError`.

### Fully exposed (no prior shape check shields `shape[-2]`)

| Function | File | Line |
|----------|------|------|
| `kabsch` | `src/kabsch_horn/numpy/kabsch_svd_nd.py` | 26 |
| `kabsch_umeyama` | `src/kabsch_horn/numpy/kabsch_svd_nd.py` | 133 |
| `kabsch` | `src/kabsch_horn/pytorch/kabsch_svd_nd.py` | 143 |
| `kabsch_umeyama` | `src/kabsch_horn/pytorch/kabsch_svd_nd.py` | 252 |
| `kabsch` | `src/kabsch_horn/jax/kabsch_svd_nd.py` | 118 |
| `kabsch_umeyama` | `src/kabsch_horn/jax/kabsch_svd_nd.py` | 218 |
| `kabsch` | `src/kabsch_horn/tensorflow/kabsch_svd_nd.py` | 98 |
| `kabsch_umeyama` | `src/kabsch_horn/tensorflow/kabsch_svd_nd.py` | 200 |

### Partially shielded (a `shape[-1] != 3` check appears first, but a 1D input of length 3 still falls through)

| Function | File | Line |
|----------|------|------|
| `horn` | `src/kabsch_horn/numpy/horn_quat_3d.py` | 83 |
| `horn_with_scale` | `src/kabsch_horn/numpy/horn_quat_3d.py` | 168 |
| `horn` | `src/kabsch_horn/jax/horn_quat_3d.py` | 145 |
| `horn_with_scale` | `src/kabsch_horn/jax/horn_quat_3d.py` | 216 |
| `horn` | `src/kabsch_horn/tensorflow/horn_quat_3d.py` | 74 |
| `horn_with_scale` | `src/kabsch_horn/tensorflow/horn_quat_3d.py` | 202 |
| `kabsch` | `src/kabsch_horn/mlx/kabsch_svd_nd.py` | 105 |
| `kabsch_umeyama` | `src/kabsch_horn/mlx/kabsch_svd_nd.py` | 224 |
| `horn` | `src/kabsch_horn/mlx/horn_quat_3d.py` | 68 |
| `horn_with_scale` | `src/kabsch_horn/mlx/horn_quat_3d.py` | 179 |

### Not directly affected (delegates to an affected function)

`kabsch_rmsd` and `kabsch_umeyama_rmsd` in PyTorch, JAX, TensorFlow, and MLX all delegate to `kabsch`/`kabsch_umeyama`, so they inherit the fix automatically.

## Problem

A 1D array/tensor is not a valid point cloud input. The API expects `[..., N, D]` (at least 2D). But the first dimensional validation in every function is:

```python
if P.shape[-2] < 2:
    raise ValueError("At least 2 points are required for alignment")
```

When `P` is 1D, `P.shape[-2]` raises `IndexError: tuple index out of range` because there is no second-to-last dimension.

## Fix

Add an ndim/rank check as the first validation in every affected function, before any `shape[-2]` access:

```python
if P.ndim < 2:
    raise ValueError(
        f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
    )
```

For TensorFlow, use the equivalent `tf.rank` or static `.shape.rank` check.

## Success criteria

- All 18 listed functions reject 1D input with `ValueError` (not `IndexError`)
- Error message includes the actual shape for debugging
- Existing tests continue to pass
- Add a parametrized test across all frameworks that passes a 1D input and asserts `ValueError`

## Found by

NumPy framework audit (input validation), extended to all frameworks.

hunter-heidenreich:
## Summary
- Adds `ndim < 2` / `rank < 2` guard as the first dimensional validation in all 18 public functions across NumPy, PyTorch, JAX, TensorFlow, and MLX
- 1D inputs now raise `ValueError` with the actual shape instead of an opaque `IndexError: tuple index out of range`
- TensorFlow functions also get a dynamic `tf.debugging.assert_rank_at_least` for graph-mode coverage

Closes #153

## Test plan
- [x] New `TestRejects1DInput` class in `test_error_handling.py` -- parametrized across all 4 algorithms and all framework adapters
- [x] All existing tests pass (`uv run pytest tests/`)

### DIFF
--- a/src/kabsch_horn/jax/horn_quat_3d.py
+++ b/src/kabsch_horn/jax/horn_quat_3d.py
@@ -140,6 +140,10 @@ def horn(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
     if P.shape[-2] < 2:
@@ -216,6 +220,10 @@ def horn_with_scale(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
--- a/src/kabsch_horn/jax/kabsch_svd_nd.py
+++ b/src/kabsch_horn/jax/kabsch_svd_nd.py
@@ -115,6 +115,10 @@ def kabsch(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-2] < 2:
         raise ValueError("At least 2 points are required for alignment")
 
@@ -220,6 +224,10 @@ def kabsch_umeyama(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     if P.shape[-2] < 2:
         raise ValueError("At least 2 points are required for alignment")
--- a/src/kabsch_horn/mlx/horn_quat_3d.py
+++ b/src/kabsch_horn/mlx/horn_quat_3d.py
@@ -73,6 +73,10 @@ def horn(P: mx.array, Q: mx.array) -> tuple[mx.array, mx.array, mx.array]:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     _warn_if_float64(P, Q)
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
@@ -191,6 +195,10 @@ def horn_with_scale(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     _warn_if_float64(P, Q)
     if P.shape[-1] != 3:
--- a/src/kabsch_horn/mlx/kabsch_svd_nd.py
+++ b/src/kabsch_horn/mlx/kabsch_svd_nd.py
@@ -118,6 +118,10 @@ def kabsch(P: mx.array, Q: mx.array) -> tuple[mx.array, mx.array, mx.array]:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
 
     if P.shape[-1] != 3:
         raise ValueError(
@@ -244,6 +248,10 @@ def kabsch_umeyama(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
 
     if P.shape[-1] != 3:
--- a/src/kabsch_horn/numpy/horn_quat_3d.py
+++ b/src/kabsch_horn/numpy/horn_quat_3d.py
@@ -78,6 +78,10 @@ def horn(P: np.ndarray, Q: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarr
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
     if P.shape[-2] < 2:
@@ -168,6 +172,10 @@ def horn_with_scale(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
--- a/src/kabsch_horn/numpy/kabsch_svd_nd.py
+++ b/src/kabsch_horn/numpy/kabsch_svd_nd.py
@@ -23,6 +23,10 @@ def kabsch(P: np.ndarray, Q: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.nda
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-2] < 2:
         raise ValueError("At least 2 points are required for alignment")
 
@@ -135,6 +139,10 @@ def kabsch_umeyama(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     if P.shape[-2] < 2:
         raise ValueError("At least 2 points are required for alignment")
--- a/src/kabsch_horn/pytorch/horn_quat_3d.py
+++ b/src/kabsch_horn/pytorch/horn_quat_3d.py
@@ -63,6 +63,10 @@ def horn(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
     if P.shape[-2] < 2:
@@ -191,6 +195,10 @@ def horn_with_scale(
     if P.shape != Q.shape:
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
+        )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
         )
     if P.shape[-1] != 3:
         raise ValueError("Horn's method is strictly for 3D point clouds")
--- a/src/kabsch_horn/pytorch/kabsch_svd_nd.py
+++ b/src/kabsch_horn/pytorch/kabsch_svd_nd.py
@@ -140,6 +140,10 @@ def kabsch(
         raise ValueError(
             f"P and Q must have the same shape, got {P.shape} vs {Q.shape}"
         )
+    if P.ndim < 2:
+        raise ValueError(
+            f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
+        )
     if P.shape[-2] < 2:
         raise ValueError("At least 2 points are required for alignment")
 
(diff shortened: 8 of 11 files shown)


hunter-heidenreich:
## API Consistency Review -- PR #171: `ndim < 2` validation guard

I audited all 10 source files and the test file to verify cross-framework consistency of the new `ndim < 2` check. Here is a detailed breakdown.

### Validation ordering by function

**Kabsch functions** (`kabsch_svd_nd.py` -- `kabsch` and `kabsch_umeyama`):

| Order | NumPy | PyTorch | JAX | TensorFlow | MLX |
|-------|-------|---------|-----|------------|-----|
| 1 | shape equality | shape equality | shape equality | shape equality (static + dynamic) | shape equality |
| 2 | ndim < 2 | ndim < 2 | ndim < 2 | rank < 2 (static + dynamic) | ndim < 2 |
| 3 | shape[-2] < 2 | shape[-2] < 2 | shape[-2] < 2 | shape[-2] < 2 (static + dynamic) | shape[-1] != 3 |
| 4 | -- | -- | -- | -- | shape[-2] < 2 |

**Horn functions** (`horn_quat_3d.py` -- `horn` and `horn_with_scale`):

| Order | NumPy | PyTorch | JAX | TensorFlow | MLX |
|-------|-------|---------|-----|------------|-----|
| 1 | shape equality | shape equality | shape equality | shape equality (static + dynamic) | shape equality |
| 2 | ndim < 2 | ndim < 2 | ndim < 2 | rank < 2 (static + dynamic) | ndim < 2 |
| 3 | shape[-1] != 3 | shape[-1] != 3 | shape[-1] != 3 | shape[-1] != 3 (static + dynamic) | `_warn_if_float64` |
| 4 | shape[-2] < 2 | shape[-2] < 2 | shape[-2] < 2 | shape[-2] < 2 (static + dynamic) | shape[-1] != 3 |
| 5 | -- | -- | -- | -- | shape[-2] < 2 |

All five frameworks follow the same logical ordering. Good.

### Error message wording

All frameworks use the identical message string for the ndim check:

```
f"Input must be at least 2D with shape [..., N, D], got shape {P.shape}"
```

TensorFlow's dynamic check uses a slightly shortened version (no f-string with shape):
```
"Input must be at least 2D with shape [..., N, D]"
```

This is expected and correct -- TF's `tf.debugging.assert_rank_at_least` cannot interpolate the shape into the message at graph-construction time.

### TensorFlow dual-check pattern

TensorFlow correctly uses both static and dynamic checks for all three validation guards:

- **Shape equality**: `P.shape != Q.shape` (static) + `tf.debugging.assert_equal` (dynamic)
- **ndim/rank**: `P.shape.rank is not None and P.shape.rank < 2` (static) + `tf.debugging.assert_rank_at_least` (dynamic)
- **shape[-2] < 2**: `P.shape[-2] is not None and P.shape[-2] < 2` (static) + `tf.debugging.assert_greater_equal` (dynamic)

The `P.shape.rank is not None` guard is correct -- it avoids errors when TF shapes are unknown at graph construction time. Consistent across `kabsch`, `kabsch_umeyama`, `horn`, and `horn_with_scale`.

### Exception types

- NumPy, PyTorch, JAX, MLX: all raise `ValueError` directly.
- TensorFlow: static checks raise `ValueError`; dynamic checks raise `tf.errors.InvalidArgumentError`.

All consistent with the existing pattern for the other checks in this codebase.

### MLX: `_warn_if_float64` ordering

There is an inconsistency in where `_warn_if_float64` is called relative to the ndim check across MLX Kabsch vs Horn:

**MLX Kabsch** (`kabsch_svd_nd.py`):
```
shape equality -> ndim < 2 -> shape[-1] != 3 -> shape[-2] < 2 -> _warn_if_float64
```

**MLX Horn** (`horn_quat_3d.py`):
```
shape equality -> ndim < 2 -> _warn_if_float64 -> shape[-1] != 3 -> shape[-2] < 2
```

In Horn, `_warn_if_float64` is called *between* the ndim check and the `shape[-1] != 3` check (lines 80-82 of `mlx/horn_quat_3d.py`). In Kabsch, it comes *after* all validation checks. This is a pre-existing inconsistency that predates this PR, so I am noting it as a nitpick rather than blocking.

### `kabsch_rmsd` / `kabsch_umeyama_rmsd` inheritance

All four autodiff frameworks (PyTorch, JAX, TensorFlow, MLX) define these as thin wrappers that delegate directly to `kabsch()` / `kabsch_umeyama()`:

```python
def kabsch_rmsd(P, Q):
    _R, _t, rmsd = kabsch(P, Q)
    return rmsd
```

The ndim check is inherited automatically. No additional guards needed. Confirmed for all four frameworks.

NumPy has no rmsd wrappers (forward-only), which is correct per the architecture.

### Test coverage: `TestRejects1DInput`

The test at line 204 uses:
```python
with pytest.raises((ValueError, Exception), match=r"2D"):
```

This catch tuple `(ValueError, Exception)` is overly broad -- `Exception` is a superclass of nearly everything. Since all frameworks now consistently raise `ValueError` for this check, this should be tightened.

---

### Issues

#### [RECOMMENDED] Tighten exception type in `test_1d_input_raises_value_error`

**File**: `tests/test_error_handling.py`, line 204

The test uses `pytest.raises((ValueError, Exception), match=r"2D")`. Since every framework raises `ValueError` for the ndim check (TF's static check fires first, raising `ValueError` before the dynamic `InvalidArgumentError` could trigger), this should be narrowed to just `ValueError`:

```python
with pytest.raises(ValueError, match=r"2D"):
```

Using `Exception` in the tuple defeats the purpose of asserting the correct exception type. If a framework accidentally raised `RuntimeError` or `IndexError` with "2D" in the message, this test would still pass, masking a real bug.

#### [NITPICK] MLX Horn: `_warn_if_float64` placement inconsistency

In `mlx/horn_quat_3d.py`, `_warn_if_float64` is called between the ndim check and the `shape[-1] != 3` check (line 80). In `mlx/kabsch_svd_nd.py`, it is called after all validation (line 133). This means a user passing a float64 1D array to MLX Horn would see a deprecation warning before the `ValueError`, while the same input to MLX Kabsch would get a clean `ValueError` with no warning.

This predates this PR, so it is not a blocker, but worth noting for a future cleanup.

### Verdict

**APPROVED** -- The ndim < 2 guard is correctly and consistently implemented across all 5 frameworks and all 4 public functions. The validation ordering matches the expected pattern (shape equality, ndim, dimension-specific checks, minimum points). Error messages are identical. TensorFlow correctly uses both static and dynamic checks. The rmsd wrappers inherit the check through delegation. The one recommended change (tightening the test exception type) is non-blocking but worth addressing.

hunter-heidenreich:
## TensorFlow Review

The TF validation guards are well-implemented and follow TF-idiomatic patterns. The dual static/dynamic check approach is the correct way to handle TensorFlow's eager + graph mode duality.

### Issues

#### [RECOMMENDED] `kabsch` and `kabsch_umeyama` lack `tf.convert_to_tensor` before validation

The Horn functions (`horn`, `horn_with_scale`) correctly call `tf.convert_to_tensor(P)` / `tf.convert_to_tensor(Q)` before any shape checks. The Kabsch functions (`kabsch`, `kabsch_umeyama`) do not. This means passing a raw Python list or NumPy array to `kabsch()` would crash on `P.shape.rank` (lists have no `.shape`, and `np.ndarray.shape` is a plain tuple with no `.rank` attribute).

This is a pre-existing issue, not introduced by this PR, but the new rank checks make it more visible -- the first thing that touches `P` is now `P.shape != Q.shape` (which would also fail on a list), then `P.shape.rank` (which would fail on an ndarray). Consider adding `tf.convert_to_tensor` at the top of `kabsch` and `kabsch_umeyama` for consistency with the Horn functions. This can be a follow-up.

#### [NITPICK] `P.shape.rank` vs `P.shape.ndims`

Both `TensorShape.rank` and `TensorShape.ndims` are equivalent (one is an alias for the other). The choice of `.rank` is fine and arguably reads more clearly. Just noting for completeness that either would work identically.

### Analysis of the Pattern

**Static check (`P.shape.rank is not None and P.shape.rank < 2`)**: Correct. `TensorShape.rank` returns `None` for unknown-rank tensors (e.g., inside `@tf.function` with unspecified input signatures). The `is not None` guard ensures this does not crash in graph mode. Raises `ValueError`, matching the other frameworks (PyTorch, JAX, NumPy, MLX all raise `ValueError`).

**Dynamic check (`tf.debugging.assert_rank_at_least`)**: Correct. This is the standard TF op for runtime rank assertions. It covers graph-mode execution where static shape info is unavailable. It raises `tf.errors.InvalidArgumentError` (not `ValueError`), which is the expected TF behavior for dynamic assertions. This is acceptable -- users in graph mode expect TF error types.

**Exception type asymmetry**: The static path raises `ValueError` and the dynamic path raises `tf.errors.InvalidArgumentError`. This is consistent with how the existing `shape[-2] < 2` and `shape != Q.shape` checks already work in these files (same dual pattern), so there is no new inconsistency.

**Ordering**: The rank check comes before the `P.shape[-2]` check in all four functions. This is correct -- accessing `P.shape[-2]` on a 0D or 1D tensor would raise an `IndexError` without the rank guard.

**No interference with `@tf.custom_gradient`**: The rank checks are in `kabsch`, `kabsch_umeyama`, `horn`, and `horn_with_scale` -- the caller functions. The `@tf.custom_gradient` decorators are on `safe_svd` and `call_safe_eigh`, which are internal functions called after validation. No interference.

**Horn functions have `tf.convert_to_tensor` before checks**: Confirmed. `horn` and `horn_with_scale` both call `tf.convert_to_tensor(P)` at lines 59-60 and 200-201, before any shape access. This ensures `.shape.rank` is always available on a proper `tf.Tensor`.

### Verdict
APPROVED

hunter-heidenreich:
## Test Quality Review

`TestRejects1DInput` is a well-structured addition that addresses issue #153. The parametrization over `ALGORITHMS x frameworks` produces good coverage across all five framework backends and all four algorithm variants, matching the 18 affected functions listed in the issue. The AAA structure is clean and easy to follow, and the test name accurately describes the behavior under test.

That said, there is one blocking issue with the assertion that would allow the original bug to pass the test silently.

### Issues

#### [BLOCKING] `pytest.raises((ValueError, Exception))` renders the assertion toothless

```python
with pytest.raises((ValueError, Exception), match=r"2D"):
    func(P, Q)
```

`Exception` is the base class of nearly all built-in exceptions. This tuple means the test will pass for **any** exception whose message contains "2D" -- including `IndexError`, which is the exact bug this test is supposed to catch. If a framework regresses and starts raising `IndexError("tuple index out of range")` again, the test would only fail because of the `match` string, not because of the exception type.

The issue's success criteria explicitly state: *"All 18 listed functions reject 1D input with `ValueError` (not `IndexError`)"*. The assertion should be:

```python
with pytest.raises(ValueError, match=r"2D"):
    func(P, Q)
```

If some frameworks raise a different exception type for valid reasons (like JAX's `TypeError` or TF's `InvalidArgumentError`), use `adapter.mismatch_exception_type` or define a new adapter property (e.g., `adapter.rank_exception_type`) -- exactly as the existing `test_raises_error_when_point_counts_differ` test does. But catching bare `Exception` defeats the purpose of the test.

#### [RECOMMENDED] Test the "partially shielded" Horn path with a length-3 input

The issue specifically calls out that Horn functions have a `shape[-1] != 3` guard that fires first -- but a 1D input of **exactly length 3** slips past it and hits `shape[-2]`. The current test uses `np.array([1.0, 2.0, 3.0])`, which is length 3, so it does exercise this path. However, this is load-bearing and not documented in the test. A brief comment explaining that the length-3 input is intentional (to bypass the Horn `D != 3` guard) would make the test's intent clearer and prevent someone from changing the fixture data without understanding why it matters.

#### [RECOMMENDED] Consider a 0D (scalar) input test

A scalar input (`np.float64(1.0)`) has `ndim == 0` and would also fail at `shape[-2]` with an `IndexError` before the fix. The `ndim < 2` guard covers this, but there is no test verifying it. Adding a scalar case (even as a second parametrize value alongside the 1D array) would strengthen the boundary testing:

```python
@pytest.mark.parametrize("bad_input", [
    pytest.param(np.array([1.0, 2.0, 3.0]), id="1D"),
    pytest.param(np.float64(1.0), id="0D"),
])
```

#### [NITPICK] `match=r"2D"` could be slightly more specific

The regex `r"2D"` is short enough that it could match unintended substrings in error messages (e.g., "float32D" -- unlikely but not impossible). A slightly more anchored pattern like `r"at least 2D"` (matching the actual error message text) would be more robust. Low priority since the current messages are controlled by this project.

### Verdict

**CHANGES REQUESTED** -- the `(ValueError, Exception)` catch tuple is a correctness issue that must be fixed before merge. The test as written cannot distinguish between the bug (`IndexError`) and the fix (`ValueError`), which is the entire point of this test class.

hunter-heidenreich:
## Code Style Review

All 10 source files and `tests/test_error_handling.py` pass `ruff check` with no issues. Error messages are consistent, line lengths are within the 88-char limit, f-strings are correct, and type hints on the test are complete.

### Issues

#### [NITPICK] Test docstring uses "not X" negation pattern

The class docstring reads:

```python
"""1D array inputs must raise ValueError, not IndexError."""
```

Per `.github/copilot-instructions.md`, avoid negation/contrastive reframes ("not X, but Y"). A simple reword:

```python
"""1D array inputs must raise ValueError with a clear message."""
```

#### [NITPICK] `pytest.raises` exception tuple is redundant

```python
with pytest.raises((ValueError, Exception), match=r"2D"):
```

`Exception` is a superclass of `ValueError`, so `(ValueError, Exception)` is equivalent to just `Exception`. If the intent is to catch `ValueError` specifically but also tolerate framework-specific exceptions (e.g., TF's `InvalidArgumentError`), this works but is a bit misleading. Two options:

1. If all frameworks are expected to raise `ValueError`: use `pytest.raises(ValueError, match=r"2D")`.
2. If TF might raise `tf.errors.InvalidArgumentError` at runtime: use `pytest.raises((ValueError, tf.errors.InvalidArgumentError), match=r"2D")` or keep `Exception` but drop `ValueError` from the tuple since it is already covered.

This is purely a readability concern -- the test will work correctly either way.

#### [NITPICK] TF dynamic assertion message omits shape detail

The static `ValueError` messages across all 20 occurrences consistently use:

```
Input must be at least 2D with shape [..., N, D], got shape {P.shape}
```

The 4 TF `tf.debugging.assert_rank_at_least` calls use a shortened form:

```
Input must be at least 2D with shape [..., N, D]
```

This is acceptable since `tf.debugging` appends its own shape info, but noting it for completeness.

### Verdict
APPROVED

